Issue #9, opened 2014-07-11
IFIleSystem should use UTC dates \ regular 


pranavkm:
https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.AspNet.FileSystems/PhysicalFileSystem.cs#L207-L210

LastWriteTimeUtc

PR #25, opened 2014-12-05
LastModified date should use a DateTimeOffset, and reflect UTC

mj1856:
Addresses #9


mj1856:
Not sure why this is still open.  It looks like these were already merged back in December.  Perhaps they were cherry-picked?

Anyway, closing the PR.

PR #31, opened 2014-12-17
Handling the directory rename case for file system watcher

Praburaj:
Today when a trigger is registered for items under a directory and when the directory is renamed FSW only fires an event for the directory change, but not for the files/sub-directories under it. So if there are triggeres setup for sub-items, they they trigger on folder rename. This fix addresses the issue, by handling a directory change event and raising the trigger for the file system entries under the renamed directory.

@davidfowl @pranavkm


Tratcher:
:shipit:

PR #32, opened 2015-01-04
Modify FileChangeTrigger to additionally check for files on disk

pranavkm:
In the event the FileSystemWatcher fails to fire file system notifications
events.

### DIFF
--- a/src/Microsoft.AspNet.FileSystems/Implementation/FileChangeTrigger.cs
+++ b/src/Microsoft.AspNet.FileSystems/Implementation/FileChangeTrigger.cs
@@ -2,71 +2,56 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Threading;
-using System.Threading.Tasks;
-using System.Text.RegularExpressions;
-using Microsoft.Framework.Expiration.Interfaces;
 
 namespace Microsoft.AspNet.FileSystems
 {
-    internal class FileChangeTrigger : IExpirationTrigger
+    internal class FileChangeTrigger : FileExpirationTriggerBase
     {
-        private CancellationTokenSource TokenSource { get; set; } = new CancellationTokenSource();
+        private static readonly TimeSpan FileCheckInterval = TimeSpan.FromSeconds(3);
+        private readonly IFileSystem _fileSystem;
+        private DateTime _lastRefreshTime;
+        private DateTimeOffset _lastModified;
 
-        public FileChangeTrigger(string pattern)
+        public FileChangeTrigger(IFileSystem fileSystem, string filePath)
         {
-            Pattern = pattern;
+            _fileSystem = fileSystem;
+            FilePath = filePath;
+            UpdateFileInfo();
         }
 
-        public string Pattern { get; private set; }
+        public string FilePath { get; }
 
-        private Regex _searchRegex;
-        private Regex SearchRegex
+        public override bool IsExpired
         {
-            get
-            {
-                if (_searchRegex == null)
-                {
-                    // Perf: Compile this as this may be used multiple times.
-                    _searchRegex = new Regex('^' + Pattern + '$', RegexOptions.IgnoreCase | RegexOptions.ExplicitCapture);
-                }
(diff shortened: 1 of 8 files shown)


pranavkm:
cc @Praburaj \ @Tratcher  \ @rynowak per our offline discussion

src/Microsoft.AspNet.FileSystems/Implementation/FileExpirationTriggerBase.cs
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Framework.Expiration.Interfaces;
+
+namespace Microsoft.AspNet.FileSystems
+{
+    public abstract class FileExpirationTriggerBase : IExpirationTrigger
```
Tratcher: Internal?

src/Microsoft.AspNet.FileSystems/Implementation/WildcardFileChangeTrigger.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text.RegularExpressions;
+
+namespace Microsoft.AspNet.FileSystems
+{
+    internal class WildcardFileChangeTrigger : FileExpirationTriggerBase
```
Tratcher: FilePatternExpirationTrigger?

src/Microsoft.AspNet.FileSystems/Implementation/FileChangeTrigger.cs
```diff
@@ -2,71 +2,56 @@
...
-                }
-
-                return _searchRegex;
-            }
+            get { return base.IsExpired || IsFileChangedInFileSystem(); }
         }
 
-        public bool ActiveExpirationCallbacks
+        // for unit testing.
+        protected virtual DateTime UtcNow
```
Tratcher: Consider using the ISystemClock pattern we use elsewhere for testing. See MemoryCache. It limits the number of custom test classes you need to create.

src/Microsoft.AspNet.FileSystems/Implementation/FileChangeTrigger.cs
```diff
@@ -2,71 +2,56 @@
...
-        {
-            return TokenSource.Token.Register(callback, state);
-        }
+            var previousLastModified = _lastModified;
+            UpdateFileInfo();
 
-        public bool IsMatch(string relativePath)
-        {
-            return SearchRegex.IsMatch(relativePath);
+            return previousLastModified != _lastModified;
```
Tratcher: LessThan vs NotEqualTo?

src/Microsoft.AspNet.FileSystems/Implementation/PhysicalFileSystemWatcher.cs
```diff
@@ -35,12 +34,16 @@ internal PhysicalFileSystemWatcher(string root)
         internal IExpirationTrigger CreateFileChangeTrigger(string filter)
         {
             filter = NormalizeFilter(filter);
-            var pattern = WildcardToRegexPattern(filter);
+            var isWildCardSearch = IsWildCardSearch(filter);
```
Praburaj: Can't we just implement checking `if(wildcard)` within the `FileChangeTrigger` itself so that it becomes an internal implementation detail rather than having separate trigger classes for wild card and regular file paths. Today files not monitored periodically for patterns is just a drawback which may eventually need to be fixed to provide guarantee. When that happens pattern / plain file name are not 2 different things. I feel that will also greatly simply the code changes needed to achieve this.

pranavkm:
Sorry should've closed this PR. I closed the corresponding PR in Mvc - https://github.com/aspnet/Mvc/pull/1727 - since it looks like the changes to use the file watcher are a bit more involved and it is unlikely I'll have a chance to look at it now. We'll revive this once we bring that back.

PR #33, opened 2015-01-13
Fixing a test failure on Win7

Praburaj:
This test specifically tests triggers with regular expressions where it looks for changes in root directrory in one trigger and another trigger monitoring any change in any sub-directory (recursive).

On Win7 when a file is modified we see a directory changed event is triggered on the parent directory automatically. This results in a flaky run on Win7. So making the test bit more restrictive to fix flakiness on Win7.


victorhurdugaci:
The test change looks good to me but I am wondering if there is a real world scenario in which an app could have a different behavior on Win7 because of this.

Praburaj:
Raising more events in general will not harm the app functionally. However it is possible that app will have to refresh the file content as the trigger is raised.

victorhurdugaci:
ok, :shipit: then

Praburaj:
This change is already in.

Issue #4, opened 2014-03-09
Rename IFileSystem to something more web specific


davidfowl:
Sounds too general at the moment and it's really only about reading files from under the application root.

Tratcher:
*Provider
Ms.Aspnet.FileProviders
IFileProvider
PhysicalFileProvider
EmbeddedFileProvider

Tratcher:
@davidfowl @Eilon do you also want to rename the repo?

Eilon:
Renaming the repo :frowning: Maybe eventually. GitHub does automatically do redirects for such renames but I'm not sure if it redirects for scenarios like cloning, etc. I'm fine waiting for some future time where we'd rename this along with other stuff...

Tratcher:
@Eilon  Github is now gracefully redirecting for repo renames. See http://github.com/aspnet/kvm

Eilon:
Yeah I just saw that earlier... I'm fine with renaming it. Just let me know when (because we have to update Universe, CIs, etc.).

Tratcher:
Whenever you're ready. It won't impact my MVC fixes.

Eilon:
Done.

PR #35, opened 2015-01-20
Rename *FileSystem to *FileProvider.

Tratcher:
#4 
This will also require changes in Hosting, StaticFiles, and MVC.


src/Microsoft.AspNet.FileProviders/project.json
```diff
@@ -1,8 +1,8 @@
 {
     "version": "1.0.0-*",
-    "description": "Implementation for ASP.NET 5 File System abstractions.",
+    "description": "Implementation for ASP.NET 5 file provider abstractions.",
```
Eilon: Since we're here, change "for" to "of"?

Eilon:
:shipit:

FileProvider.sln
```diff
@@ -1,22 +1,25 @@
...
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.FileSystems.Tests", "test\Microsoft.AspNet.FileSystems.Tests\Microsoft.AspNet.FileSystems.Tests.kproj", "{66FE5FDF-BBF9-4573-A7B7-53551731C0F9}"
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.FileProviders.Tests", "test\Microsoft.AspNet.FileProviders.Tests\Microsoft.AspNet.FileProviders.Tests.kproj", "{66FE5FDF-BBF9-4573-A7B7-53551731C0F9}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{1A060559-74DD-4B5A-BBA1-E8A441E729C1}"
 	ProjectSection(SolutionItems) = preProject
 		global.json = global.json
 	EndProjectSection
+	ProjectSection(FolderGlobals) = preProject
+		global_1json__JSONSchema = http://json.schemastore.org/global
```
davidfowl: What is this crap?

Tratcher: Ask VS, they keep adding it.

dougbu: FYI @ToddGrun is working on making this addition of the default schema stop.
Suggest reverting these lines.  Otherwise you'll be locking in the current default.

Tratcher: It has already worked its way into many of the sln and kproj files. Once the issue is fixed in VS we'll want to do a general scrub to remove them.

Issue #51, opened 2015-01-22
Need to make sure that we don't return hidden files


glennc:
Files starting with . or with the hidden attribute for example.

davidfowl:
Why?

Praburaj:
This was out of the security review discussion.

PR #52, opened 2015-01-22
Refactory #1

troydai:
I'd like to use this PR to follow up the comments regarding to Microsoft.Framework.FileSystemGlobbing code.


src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -9,10 +10,14 @@ namespace Microsoft.Framework.FileSystemGlobbing.Infrastructure
 {
     public class MatcherContext
     {
+        private FrameData _frame;
+        private Stack<FrameData> _frameStack = new Stack<FrameData>();
```
pranavkm: `readonly`

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -65,194 +72,196 @@ public PatternMatchingResult Execute()
 
         private void DoPredicting()
         {
-            if (Frame.Stage != Stage.Predicting)
+            if (_frame.Stage != Stage.Predicting)
             {
                 return;
             }
+
             //foreach (var patternContext in IncludePatternContexts)
```
pranavkm: Maybe remove this?

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -65,194 +72,196 @@ public PatternMatchingResult Execute()
...
             {
                 PopFrame();
                 return;
             }
 
-            PushFrame(Stage.Predicting, Frame.ActualDirectoryEnumerator.Current);
+            PushFrame(Stage.Predicting, _frame.ActualDirectoryEnumerator.Current);
         }
 
+        private bool TestByPatterns<T>(T target, Func<PatternContextBase, T, bool> testPattern)
```
pranavkm: `TFileSystemInfo`

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -65,194 +72,196 @@ public PatternMatchingResult Execute()
...
+            _frameStack.Push(_frame);
+
+            _frame = new FrameData
             {
                 Stage = stage,
                 DirectoryInfo = directoryInfo,
                 RelativePath = relativePath,
             };
+
             foreach (var x in IncludePatternContexts)
```
pranavkm: `var context in ..`

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: yeah.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: I'll use the Linq instead.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: yup.

troydai:
@loudej @davidfowl @Praburaj 
In latest commit, I clean up the codes in `MatchContext`. I recognize that the `MatchContext.Execute` is essentially a __Pre-Order, Non-recursive, tree traversal__. Base on this assumption I made following changes:

1. Remove the guardian `FrameData` which is used to determine the termination of the loop. Instead, it now directly detects the size of the stack.
2. Remove unused stages. Instead, there will be only two stages for a node, *visited* or *not visited*. Children enumerating will happen in when the current node has been visited.
3. The original implementation of iterating the siblings requires three properties on the `FrameData` a boolean, a list and an enumerator. I change them to an array and an index to facilitate the logic. It not only reducing the number of properties but also make the logic more readable. In previous review it clear the enumerator brings confusions.

There are also other minor changes for cleaning up the codes. They were not listed.

troydai:
@Eilon @pranavkm @davidfowl @loudej

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -3,275 +3,180 @@
...
+using System.Linq;
 using Microsoft.Framework.FileSystemGlobbing.Abstractions;
 
 namespace Microsoft.Framework.FileSystemGlobbing.Infrastructure
 {
     public class MatcherContext
     {
+        private DirectoryInfoBase _root;
+        private IList<PatternContextBase> _includePatternContexts;
+        private IList<PatternContextBase> _excludePatternContexts;
```
davidfowl: readonly?

troydai: will do.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -3,275 +3,180 @@
...
-                if (pattern.Contains == null)
-                {
-                    ExcludePatternContexts.Add(new PatternContextLinearExclude(this, pattern));
-                }
-                else
-                {
-                    ExcludePatternContexts.Add(new PatternContextRaggedExclude(this, pattern));
-                }
+                _excludePatternContexts.Add(pattern.Contains == null ?
+                    (PatternContextBase)new PatternContextLinearExclude(this, pattern) : new PatternContextRaggedExclude(this, pattern));
```
Praburaj: Do you need this casting?

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -3,275 +3,180 @@
...
+            PushFrame(new FrameData(_root), stack);
+            while (stack.Count > 0)
             {
-                DoPredicting();
-                DoEnumerating();
-                DoRecursion();
+                var current = stack.Peek();
+
+                DoEnumerating(current, files);
+                DoRecursion(current, stack);
```
Eilon: These two method names have been confusing me. Is it really this:
```c#
ProcessFilesInCurrentFolder(...);
ProcessSubfolders(...);
```
?

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -3,275 +3,180 @@
...
 
-            var moveNext = Frame.ActualDirectoryEnumerator.MoveNext();
-            if (moveNext == false)
+        private void PushFrame(FrameData frame, Stack<FrameData> stack)
+        {
+            if (stack.Count() > 0)
             {
-                PopFrame();
-                return;
+                frame.RelativePath = stack.Peek().RelativePath + '/' + frame.DirectoryInfo.Name;
```
Praburaj: Can `Path.Combine()` be used instead here?

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: Yeah, legacy names. I'll update.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: You're right. There are a few other places using the same pattern. I'm going to sweep them together.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: Yes we need. The `? :` operator can't deduce the correct return type when the two variables on the both side of `:` are of different type even through they share the same base type. The operator simply don't climb the hierarchy on both type.

Praburaj: why not just use a `if` `else` then?

lodejard: Yeah, ternary operator seems like it's less clear even if it is more compact

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
lodejard: It's not really quite that... DoEnumerating is when you're in the stage of deciding what are the contents of the current focus. DoRecursion is doing the actual processing of contents of that focus, and recursing is really just a part of that.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
lodejard: Path.Combine does more than just combination, and doesn't use the '/' character, so we'd also need decide if we want to use '/' for the output of this algorithm or let it go back and forth

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: Yeah, it seems to be. I'll reverse the change to an `if-else` structure to gain readability.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: Suggestions: 

```
VisitCurrent
PushChildren
```

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
troydai: I think using `Path.Combine` is a safer approach. Another though is that we don't need to compose the `relativePath` here. Instead we can use a `List` to store the segments while when a relative path in the form of `string` is required, then the consumer decide how to concatenate them.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -3,275 +3,180 @@
...
         {
-            Files = new List<string>();
+            var files = new List<string>();
 
-            Frame.Stage = Stage.Complete;
-            PushFrame(Stage.Predicting, DirectoryInfo);
+            // Walk the file system in Non-recursive, pre order, tree traversal
 
-            while (Frame.Stage != Stage.Complete)
+            var stack = new Stack<FrameData>();
```
lodejard: The FrameStack as member, Frame as field, and FrameData as value type were to avoid allocations for working structures. Should switch back to that technique.

troydai:
@davidfowl @loudej @Eilon @Praburaj @pranavkm  :watch:

src/Microsoft.Framework.FileSystemGlobbing/Abstractions/DirectoryInfoWrapper.cs
```diff
@@ -8,16 +8,16 @@ namespace Microsoft.Framework.FileSystemGlobbing.Abstractions
...
         public DirectoryInfoWrapper(DirectoryInfo directoryInfo)
         {
-            DirectoryInfo = directoryInfo;
+            _directoryInfo = directoryInfo;
         }
 
         public override IEnumerable<FileSystemInfoBase> EnumerateFileSystemInfos(string searchPattern, SearchOption searchOption)
         {
-            foreach (var fileSystemInfo in DirectoryInfo.EnumerateFileSystemInfos(searchPattern, searchOption))
+            foreach (var fileSystemInfo in _directoryInfo.EnumerateFileSystemInfos(searchPattern, searchOption))
```
Praburaj: How do we treat files starting with `.` or `hidden` or `system`. This originally came up in security discussion https://github.com/aspnet/FileSystem/issues/51 @GrabYourPitchforks @glennc

troydai: It's up to the implementation of `DirectoryInfoBase`. The consumer of `Matcher` decides what's the implementation is: https://github.com/aspnet/FileSystem/blob/filesystemglobbing.followup.1/src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs#L27

davidfowl: This is a general purpose library, not specifically used for Web. I don't think it should do any of those things by default.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -9,269 +10,176 @@ namespace Microsoft.Framework.FileSystemGlobbing.Infrastructure
...
-                        }
-                    }
-                    if (include)
-                    {
-                        if (Frame.ActualDirectories == null)
-                        {
-                            Frame.ActualDirectories = new List<DirectoryInfoBase>();
-                        }
-                        Frame.ActualDirectories.Add(directoryInfo);
+                        _files.Add(parentRelativePath == null ?
```
davidfowl: Instead of doing this in 2 places, let combine handle nulls.

src/Microsoft.Framework.FileSystemGlobbing/Infrastructure/MatcherContext.cs
```diff
@@ -9,269 +10,176 @@ namespace Microsoft.Framework.FileSystemGlobbing.Infrastructure
...
+        {
+            return string.Format("{0}/{1}", left, right);
+        }
+
+        private bool MatchPatternContexts<TFileInfoBase>(TFileInfoBase fileinfo, Func<PatternContextBase, TFileInfoBase, bool> test)
+        {
+            var found = false;
+
+            // If the given file/directory matches any including pattern, continues to next step.
+            foreach (var pattern in _includePatternContexts)
```
davidfowl: it's called `pattern` here and `context` below.  Make those consistent.

troydai:
:watch:

troydai:
@davidfowl ship it?

lodejard:
:shipit:

davidfowl:
:shipit:

PR #53, opened 2015-01-24
File provider not to return hidden files or file names starting with per...

Praburaj:
...iod

Fixes: https://github.com/aspnet/FileSystem/issues/51

@GrabYourPitchforks @glennc

### DIFF
--- a/src/Microsoft.AspNet.FileProviders/Implementation/PhysicalFilesWatcher.cs
+++ b/src/Microsoft.AspNet.FileProviders/Implementation/PhysicalFilesWatcher.cs
@@ -80,6 +80,12 @@ namespace Microsoft.AspNet.FileProviders
 
         private void OnFileSystemEntryChange(string fullPath)
         {
+            var fileSystemInfo = new FileInfo(fullPath);
+            if (FileSystemInfoHelper.IsHiddenFile(fileSystemInfo))
+            {
+                return;
+            }
+
             var relativePath = fullPath.Replace(_root, string.Empty);
             if (_triggerCache.ContainsKey(relativePath))
             {
--- a/src/Microsoft.AspNet.FileProviders/Implementation/FileSystemInfoHelper.cs
+++ b/src/Microsoft.AspNet.FileProviders/Implementation/FileSystemInfoHelper.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+
+namespace Microsoft.AspNet.FileProviders
+{
+    public class FileSystemInfoHelper
+    {
+        internal static bool IsHiddenFile(FileSystemInfo fileSystemInfo)
+        {
+            if (fileSystemInfo.Name.StartsWith("."))
+            {
+                return true;
+            }
+            else if (fileSystemInfo.Exists &&
+                (fileSystemInfo.Attributes.HasFlag(FileAttributes.Hidden) ||
+                fileSystemInfo.Attributes.HasFlag(FileAttributes.System)))
+            {
+                return true;
+            }
+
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNet.FileProviders/Implementation/PhysicalFilesWatcher.cs
```diff
@@ -80,6 +80,14 @@ private void OnChanged(object sender, FileSystemEventArgs e)
 
         private void OnFileSystemEntryChange(string fullPath)
         {
+            var fileSystemInfo = new FileInfo(fullPath);
+            if (fileSystemInfo.Name.StartsWith(".") ||
+                (fileSystemInfo.Exists && fileSystemInfo.Attributes.HasFlag(FileAttributes.Hidden)))
```
Eilon: This looks like it should go in some static helper to re-use in the several places it's used.

Also, I think this needs to check for `FileAttributes.System`, which is a "super hidden" file.

Praburaj: sure. I can make it.

Praburaj: @Eilon I pushed an updated PR.

src/Microsoft.AspNet.FileProviders/Implementation/Helper.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+
+namespace Microsoft.AspNet.FileProviders
+{
+    public class Helper
```
Eilon: Never name things `Helper`. Need a more specific name for this.

src/Microsoft.AspNet.FileProviders/Implementation/Helper.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        internal static bool CanBeServed(FileSystemInfo fileSystemInfo)
```
Eilon: `CanBeServed` seems orthogonal to the use case. Isn't this just saying whether the file is hidden or not?

src/Microsoft.AspNet.FileProviders/Implementation/Helper.cs
Praburaj: How about `FileSystemInfoHelper` ?

src/Microsoft.AspNet.FileProviders/Implementation/Helper.cs
Praburaj: How about `IsReservedFile`? (Of course with return boolean negated).

Eilon: They're not *reserved*, they're *hidden*. `IsHiddenFile` would be fine.

src/Microsoft.AspNet.FileProviders/Implementation/Helper.cs
Eilon: Yeah that seems fine.

src/Microsoft.AspNet.FileProviders/Implementation/Helper.cs
Praburaj: Cool. Will update it.

Praburaj:
Updated PR.

Eilon:
Looks :shipit: to me. @GrabYourPitchforks do you want to look to see if anything is missing?

Praburaj:
@GrabYourPitchforks - do you have any comments?

Praburaj:
Changes merged : https://github.com/aspnet/FileSystem/commit/0d2d9b1913175a764faaa57dc40de3e3761797c9

PR #58, opened 2015-02-10
Fix the incorrect matched parent path segment

troydai:
## Fixing following bug: when two include pattens like following are added:
```
**\*.*
..\..\lib\*.*
```
The parent folder of `lib` is incorrectly included. That is because the `..` is matched by wildcard. 

## Add work around for CoreCLR issue https://github.com/dotnet/corefx/issues/616. 
The change hasn't been made into the drop yet. A warning test is added so that when the fix is in the build the test will be skipped so that a warning shows up in CI. It reminds us to revert the work around.

## Add more test cases to cover the scenario.

### DIFF
--- a/src/Microsoft.Framework.FileSystemGlobbing/Abstractions/DirectoryInfoWrapper.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Abstractions/DirectoryInfoWrapper.cs
@@ -14,7 +14,13 @@ namespace Microsoft.Framework.FileSystemGlobbing.Abstractions
 
         public DirectoryInfoWrapper(DirectoryInfo directoryInfo, bool isParentPath = false)
         {
-            _directoryInfo = directoryInfo;
+            // Work around for https://github.com/dotnet/corefx/issues/616: DirectoryInfo.Exists returns 
+            // incorrect results when it is cast from FileSystemInfo.
+            //
+            // Once it is fixed us following line to replace the assignment to _directoryInfo
+            // _directoryInfo = directoryInfo.FullName;
+            _directoryInfo = new DirectoryInfo(directoryInfo.FullName);
+
             _isParentPath = isParentPath;
         }
 
--- a/src/Microsoft.Framework.FileSystemGlobbing/Internal/Patterns/PatternBuilder.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Internal/Patterns/PatternBuilder.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments;
 using Microsoft.Framework.FileSystemGlobbing.Internal.PatternContexts;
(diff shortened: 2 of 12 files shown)


troydai:
@lodejard @davidfowl @Praburaj

src/Microsoft.Framework.FileSystemGlobbing/Abstractions/DirectoryInfoWrapper.cs
```diff
@@ -14,7 +14,13 @@ public class DirectoryInfoWrapper : DirectoryInfoBase
 
         public DirectoryInfoWrapper(DirectoryInfo directoryInfo, bool isParentPath = false)
         {
-            _directoryInfo = directoryInfo;
+            // Work around for https://github.com/dotnet/corefx/issues/616: DirectoryInfo.Exists returns
```
Praburaj: @troydai - the bug seems to be fixed. https://github.com/dotnet/corefx/issues/616. Do we need this fix? 

>stephentoub commented 3 days ago
Fixed in #640

Praburaj: Ah.. you seem to have mentioned that. When is the coreclr fix expected to reach us?

src/Microsoft.Framework.FileSystemGlobbing/Internal/Patterns/PatternBuilder.cs
```diff
@@ -14,9 +15,15 @@ public static class PatternBuilder
 
         public static IPattern Build(string pattern)
```
Praburaj: Is it possible to use `[NotNull]` here?

test/Microsoft.Framework.FileSystemGlobbing.Tests/TestUtility/RunWhenWhenDirectoryInfoWorksAttribute.cs
```diff
@@ -21,7 +22,7 @@ public string SkipReason
         {
             get
             {
-                return "CoreCLR DirectoryInfo incorrectly text directory existence.";
+                return "Issue: \"DirectoryInfo.Exists returns incorrect results when it is cast from FileSystemInfo.\" is fixed. Please remove the workaround in the produce code and this test case.";
```
Praburaj: Why not let the test fail when the coreclr fix is available at our end so that we automatically know. With skipping a test we need to watch if the test is skipped.

src/Microsoft.Framework.FileSystemGlobbing/Internal/Patterns/PatternBuilder.cs
troydai: What package will be introduced if using `[NotNull]`? I want to keep the dependencies to minimal.

test/Microsoft.Framework.FileSystemGlobbing.Tests/TestUtility/RunWhenWhenDirectoryInfoWorksAttribute.cs
troydai: Fail is little bit too harsh.

src/Microsoft.Framework.FileSystemGlobbing/Abstractions/DirectoryInfoWrapper.cs
troydai: I checked the latest build it's not there. @davidfowl how long does it take for the fix to appear in the CoreCLR drop?

davidfowl: No idea.

src/Microsoft.Framework.FileSystemGlobbing/Internal/Patterns/PatternBuilder.cs
Praburaj: It should be a ANI

troydai:
Ping, any further comments?

@davidfowl @lodejard @muratg

troydai:
Looks good? Going to merge soon.

PR #59, opened 2015-02-11
Process **.suffix pattern

troydai:
__Includes two changes__

* Support pattern `**.suffix*`
`**.suffix` is equivalent to `**\*.suffix`. It will be translated into two path segments: a recursive path segment and a wild-card path segment.

* Support including directory by `directory\`
The patterns end with `\` or `/` will be expand to `/**/*.*`


troydai:
@loudej @davidfowl @Praburaj @muratg

lodejard:
Couldn't the directory match be expanded to `directory/**` ?

lodejard:
Also, did the old regex based patterns support `**.suffix`?

troydai:
I should try `**`. And also `**.suffix` was supported. It is actually used in template https://github.com/aspnet/Templates/blob/master/src/StarterWeb/project.json#L43.

troydai:
@lodejard looks good?

lodejard:
:shipit:

troydai:
Super. I will rebase and merge in minutes.

Issue #60, opened 2015-02-12
Make Matcher easier to test


DamianEdwards:
`Matcher` currently is hard to use in other types in a way that makes those types testable, as you can't observe the current state of the `Matcher` (e.g. what patterns are active via properties), you can't mock it (because the members aren't virtual), and you can't substitute the `Matcher` completely without wrapping it which seems like overkill given it's in our codebase.

Perhaps we can make the members virtual, and add some protected properties to be able to see the current pattern lists (include and excludes).

davidfowl:
Can you give an example?

troydai:
@davidfowl Here's my understanding: For @DamianEdwards to unit test a component which uses the `Matcher` is difficult because the `Matcher` is none extensible. Therefore you can't mock it. 

This is a good point. Given that almost or the other components in `FileSystemGlobbing` are mockable, the `Matcher`'s testability looks odd. I believe the solution is to either introduce a `IMatcher` interface.

troydai:
Fixed in ff1f436dc5457f2aa2f520806f23d9356a09b4ac

PR #63, opened 2015-02-16
Make Matcher more testable

troydai:
Addresses issue #60 .

@DamianEdwards

### DIFF
--- a/src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs
@@ -13,19 +13,19 @@ namespace Microsoft.Framework.FileSystemGlobbing
         private IList<IPattern> _includePatterns = new List<IPattern>();
         private IList<IPattern> _excludePatterns = new List<IPattern>();
 
-        public Matcher AddInclude(string pattern)
+        public virtual Matcher AddInclude(string pattern)
         {
             _includePatterns.Add(PatternBuilder.Build(pattern));
             return this;
         }
 
-        public Matcher AddExclude(string pattern)
+        public virtual Matcher AddExclude(string pattern)
         {
             _excludePatterns.Add(PatternBuilder.Build(pattern));
             return this;
         }
 
-        public PatternMatchingResult Execute(DirectoryInfoBase directoryInfo)
+        public virtual PatternMatchingResult Execute(DirectoryInfoBase directoryInfo)
         {
             var context = new MatcherContext(_includePatterns, _excludePatterns, directoryInfo);
             return context.Execute();


src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs
```diff
@@ -10,22 +10,22 @@ namespace Microsoft.Framework.FileSystemGlobbing
 {
     public class Matcher
     {
-        private IList<IPattern> _includePatterns = new List<IPattern>();
-        private IList<IPattern> _excludePatterns = new List<IPattern>();
+        protected IList<IPattern> _includePatterns = new List<IPattern>();
+        protected IList<IPattern> _excludePatterns = new List<IPattern>();
```
DamianEdwards: Should we just go ahead and make these properties? Seems strange to have protected fields with _ prefix naming.

src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs
```diff
@@ -10,22 +10,22 @@ namespace Microsoft.Framework.FileSystemGlobbing
 {
     public class Matcher
     {
-        private IList<IPattern> _includePatterns = new List<IPattern>();
-        private IList<IPattern> _excludePatterns = new List<IPattern>();
+        protected IList<IPattern> _includePatterns { get; set; } = new List<IPattern>();
```
Praburaj: variable names to be like properties? `IncludePatterns` & `ExcludePatterns`?

troydai: Yeah. hasn't changed it.

DamianEdwards:
:shipit:

src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs
```diff
@@ -10,24 +10,24 @@ namespace Microsoft.Framework.FileSystemGlobbing
 {
     public class Matcher
     {
-        private IList<IPattern> _includePatterns = new List<IPattern>();
-        private IList<IPattern> _excludePatterns = new List<IPattern>();
+        protected IList<IPattern> IncludePatterns { get; set; } = new List<IPattern>();
+        protected IList<IPattern> ExcludePatterns { get; set; } = new List<IPattern>();
```
Eilon: Protected is not a great pattern. Why not public?

troydai: The collection is not essential to external consumption. If it is public, then we need to guard the situations in which a invalid pattern is added to the pattern collection.

Eilon: How do you guard against a derived class from adding invalid patterns? Changing a class's design *just* for a unit test is a slippery slope.

DamianEdwards: I wondered whether the properties shouldn't just be IEnumerable to enable inspection, and leave the Add/Remove as distinct APIs

-----Original Message-----

DamianEdwards: Or maybe the fields don't need to be exposed at all and we just rely on the methods being mockable

Eilon: :+1: to what @DamianEdwards said.

troydai: so, private fields. virtual methods.

Eilon:
Looks cleaner to me!

troydai:
Assume it's a :shipit: ?

Eilon:
:shipit: !

PR #64, opened 2015-02-16
Update globbing tests

troydai:
1. Add two tests covering singe file path;
2. Update assert helper to test sequence equivalence first;
3. Fix a missing field assignement in matcher.

@davidfowl


davidfowl:
:shipit:

Issue #57, opened 2015-02-06
Update Matcher and its abstractions to remove the bits not actually needed


DamianEdwards:
Parts of `Matcher` and its abstractions (e.g. `DirectoryInfoBase`) were written to support more functionality that it turns out we don't need, e.g. listing files in a directory with a pattern. We should remove the unneeded bits as they're incompatible with `IFileProvider` and not needed anyway.

DamianEdwards:
There's an implementation using `IFileProvider` of the abstractions `Matcher` uses at https://github.com/aspnet/Mvc/blob/LinkTagHelperGlobbing-1581/src/Microsoft.AspNet.Mvc.TagHelpers/Internal/FileProviderGlobbingDirectory.cs

troydai:
https://github.com/aspnet/FileSystem/commit/3a00f0b05b163120a66ea6a1ec048aee34451dbd

PR #65, opened 2015-02-16
Update interfaces

troydai:
Addressing https://github.com/aspnet/FileSystem/issues/57

@DamianEdwards anything else need to be addressed?


DamianEdwards:
Looks good to me

troydai:
Can't check in right now. Pending on __internialization__ of FileSystemGlobbing in runtime. Otherwise the test will always fail because the runtime pull in old version of FileSystemGlobbing.

troydai:
Oh, yeah. With the globbing be internalized in runtime. This PR can be merged as soon as there is a new KRE published.

DamianEdwards:
I'll wait here then...

troydai:
Changes pushed, build passed, and the package is on the volatile feed. Enjoy.

Issue #46, opened 2015-01-21
PhysicalFileWatcher shouldn't use newLocation.Replace


glennc:
In the OnRenamed event of the PhysicalFileSystemWatcher class we should just use substring and ensure that slashes are correct for all operating systems.

PR #77, opened 2015-03-17
PhysicalFileWatcher shouldn't use newLocation.Replace

Praburaj:
Fixing https://github.com/aspnet/FileSystem/issues/46

@muratg


muratg:
:shipit:

Issue #44, opened 2015-01-21
Implement error event handler on file watcher


glennc:
We listen for error events and assume that on an error * has changed.

Praburaj:
https://github.com/aspnet/FileSystem/commit/7ef9d035875354f5e885436c223536c34ff3fda8

PR #78, opened 2015-03-17
Implement error event handler on file watcher

Praburaj:
Fixes : https://github.com/aspnet/FileSystem/issues/44

@muratg


muratg:
Was this tested? If not, can we add a test case?

Otherwise :shipit:

Praburaj:
@muratg its a bit hard to simulate as it happens on these conditions per the msdn docs. https://msdn.microsoft.com/en-us/library/system.io.filesystemwatcher.error(v=vs.110).aspx

This should be a safe change as we just raise all the triggers on an error.

PR #97, opened 2015-04-06
Use HashSet to eliminiate duplicate declared folder and files

troydai:
To address the DNX issue https://github.com/aspnet/dnx/issues/1588

/cc @lodejard @loudej @davidfowl

### DIFF
--- a/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
@@ -16,8 +16,8 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
         private readonly IList<IPatternContext> _excludePatternContexts;
         private readonly IList<string> _files;
 
-        private readonly IList<LiteralPathSegment> _declaredLiteralFolderSegments = new List<LiteralPathSegment>();
-        private readonly IList<LiteralPathSegment> _declaredLiteralFileSegments = new List<LiteralPathSegment>();
+        private readonly ISet<LiteralPathSegment> _declaredLiteralFolderSegments = new HashSet<LiteralPathSegment>();
+        private readonly ISet<LiteralPathSegment> _declaredLiteralFileSegments = new HashSet<LiteralPathSegment>();
 
         private bool _declaredParentPathSegment;
         private bool _declaredWildcardPathSegment;
--- a/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
@@ -16,8 +16,8 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
         private readonly IList<IPatternContext> _excludePatternContexts;
         private readonly IList<string> _files;
 
-        private readonly ISet<LiteralPathSegment> _declaredLiteralFolderSegments = new HashSet<LiteralPathSegment>();
-        private readonly ISet<LiteralPathSegment> _declaredLiteralFileSegments = new HashSet<LiteralPathSegment>();
+        private readonly HashSet<LiteralPathSegment> _declaredLiteralFolderSegments = new HashSet<LiteralPathSegment>();
+        private readonly HashSet<LiteralPathSegment> _declaredLiteralFileSegments = new HashSet<LiteralPathSegment>();
 
         private bool _declaredParentPathSegment;
         private bool _declaredWildcardPathSegment;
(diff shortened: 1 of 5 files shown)


troydai:
I'm running a few more performance test to verify the fix.

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
```diff
@@ -17,8 +16,8 @@ public class MatcherContext
         private readonly IList<IPatternContext> _excludePatternContexts;
         private readonly IList<string> _files;
 
-        private readonly IList<LiteralPathSegment> _declaredLiteralFolderSegments = new List<LiteralPathSegment>();
-        private readonly IList<LiteralPathSegment> _declaredLiteralFileSegments = new List<LiteralPathSegment>();
+        private readonly ISet<LiteralPathSegment> _declaredLiteralFolderSegments = new HashSet<LiteralPathSegment>();
+        private readonly ISet<LiteralPathSegment> _declaredLiteralFileSegments = new HashSet<LiteralPathSegment>();
```
davidfowl: Why not just make both sides HashSet if they are assigned in place?

troydai: All right.

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
```diff
@@ -18,5 +18,18 @@ public bool Match(string value, StringComparison comparisonType)
         {
             return string.Equals(Value, value, comparisonType);
         }
+
+        public override bool Equals(object obj)
+        {
+            var other = obj as LiteralPathSegment;
+
+            return other != null &&
+                other.Value.Equals(Value, StringComparison.Ordinal);
```
Eilon: Use static `string.Equals`?

troydai:
/cc @Praburaj 

@davidfowl @Eilon Updated :watch:

davidfowl:
Why is the CI failing?

Praburaj:
Change looks good to me. Do you also need some tests for case sensitivity. Eg. "sub/source.cs" and "Sub/source.cs" etc.

troydai:
@Praburaj case-sensitive case will come in when I fix the directory case sensitive case.

troydai:
@davidfowl CI failed at package restore.

test/Microsoft.Framework.FileSystemGlobbing.Tests/FunctionalTests.cs
```diff
@@ -29,6 +29,17 @@ public void Dispose()
         }
 
         [Fact]
+        public void DuplicatePatterns()
```
davidfowl: Shouldn't this test both types of slashes?

troydai: There are tests added to test slash and backslash in the input patterns. Those tests are added when tests on *nix was fixed.

davidfowl: Use the attributes that @victorhurdugaci wrote. Have windows specific tests

troydai: Why? all the tests should pass on both Windows and *nix. I can add variation for backslash in this test. But a conditional test for Windows only shouldn't exist. We suppose to take both types of slashes.

davidfowl: Why would paths with `\` work on nix isn't that kinda weird?

troydai: Absolute path is not allowed. There are tests coverage ensuring the exceptions are thrown.

The path `\` can be very possible when a project is developed on Windows and move to *nix unless we abolish all the usage of `\`.

troydai:
:watch:

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
```diff
@@ -18,5 +18,18 @@ public bool Match(string value, StringComparison comparisonType)
...
+        {
+            var other = obj as LiteralPathSegment;
+
+            return other != null &&
+                string.Equals(other.Value, Value, StringComparison.Ordinal);
+        }
+
+        public override int GetHashCode()
+        {
+            return Value.GetHashCode();
```
victorhurdugaci: The value can be null...

troydai: :+1 I'll add guard to Value to ensure it's not null.

troydai: Will call into `base.GetHashCode()`

troydai:
CI passed. Any further comments, everyone?

:watch:

victorhurdugaci:
:shipit:

Issue #48, opened 2015-01-21
Consider EmbeddedFileSystem becoming its own NuGet package


glennc:

Praburaj:
Microsoft.AspNet.FileProviders (common stuff between providers - this contains implementation of `NoopTrigger`, `EnumerableDirectoryContents`, `NotFoundDirectoryContents` and `NotFoundFileInfo` etc..)
Microsoft.AspNet.FileProviders.Physical
Microsoft.AspNet.FileProviders.Embedded
Microsoft.AspNet.FileProviders.Interfaces

PR #100, opened 2015-04-21
Consider EmbeddedFileSystem becoming its own NuGet package

Praburaj:
Fixes: https://github.com/aspnet/FileSystem/issues/48

@Tratcher @muratg


src/Microsoft.AspNet.FileProviders.Physical/project.json
```diff
@@ -0,0 +1,20 @@
+{
+    "version": "1.0.0-*",
+    "description": "Implementation of ASP.NET 5 physical file provider abstractions.",
+    "dependencies": {
```
Tratcher: Don't you need a dependency on Interfaces here? Common gets you one transitively, but it's only a build-time dependency.  Same for Embedded.

Praburaj: Right I see that. Otherwise ones consuming these packages will have to include the file system interfaces in their projects.

Praburaj: updated.

Tratcher:
The code looks fine, though I'm not sure what is accomplished by fragmenting to this degree.

Praburaj:
This was out of the API review. My understanding is when user wants PhysicalFileSystem then we don't need to pull in Embedded one. Just a matter of organizing.

src/Microsoft.AspNet.FileProviders.Common/project.json
```diff
@@ -0,0 +1,16 @@
+{
+    "version": "1.0.0-*",
+    "description": "Implementation of ASP.NET 5 file provider abstractions.",
+    "dependencies": {
+        "Microsoft.AspNet.FileProviders.Interfaces": "1.0.0-*"
+    },
```
Eilon: Don't shared source projects need `"shared": "*.cs",` ? Or is this not shared source? The bug didn't say anything about having a package named "Microsoft.AspNet.FileProviders.Common" did it?

Tratcher: Fixed.

Tratcher:
Updated @Eilon

src/Microsoft.AspNet.FileProviders.Common/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+
+[assembly: AssemblyMetadata("Serviceable", "True")]
```
Eilon: This package doesn't really produce an assembly right? If no assembly, don't need this attribute.

src/Microsoft.AspNet.FileProviders.Common/Implementation/EnumerableDirectoryContents.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.FileProviders
+{
+    public class EnumerableDirectoryContents : IDirectoryContents
```
Eilon: I'm still a bit confused... if these shared files are shared into each assembly, won't these *public* types now be duplicated in several assemblies? Should every type in this package be internal?

src/Microsoft.AspNet.FileProviders.Embedded/project.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "1.0.0-*",
+    "description": "Implementation of ASP.NET 5 embedded file provider abstractions.",
```
Eilon: These aren't abstractions in here, right? Just "implementation"?

src/Microsoft.AspNet.FileProviders/Internal/NoopTrigger.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Expiration.Interfaces;
+
+namespace Microsoft.AspNet.FileProviders.Internal
+{
+    public class NoopTrigger : IExpirationTrigger
```
Eilon: Doesn't the new `.Common` package have this type?

Tratcher:
Updated.

Eilon:
:shipit:

Issue #101, opened 2015-04-23
Get all the unit tests running on CoreCLR


Eilon:
The main blocker right now might mean removing the dependency on Shoudly, but there could be other issues.

Eilon:
Fixed in e50b4919967d997e2f222cd367f3ceb257b75e0d.

PR #102, opened 2015-04-25
Fix 101 - make all tests run on CoreCLR

Eilon:
Removed dependency on Shouldly (which doesn't exist on CoreCLR), and
made some other small compatibility fixes.


Eilon:
Fixes #101

Eilon:
Looks like Mono tests are failing due to a difference between .NET Framework and Mono, so will investigate that later.

PR #103, opened 2015-04-26
Fix unit tests on Mono

Eilon:
OK got everything working.

- Split up some unit tests into several smaller tests that each test one thing
- Fixed some tests that were mixing physical paths and virtual paths
- Disabled one test on Mono because Mono doesn't seem to support Hidden files in the Windows file system attribute sense (no big surprise - *nix file systems don't have that concept)


Eilon:
@davidfowl

PR #110, opened 2015-06-24
Enable configure the globbing comparison type

troydai:
This change enable the consumer of globbing to configure the string comparison type. It is done by create a `Matcher` with specified `StringComparison` value.

Internally some refactory is done to achieve the goal:
1. `IPathSegment` interface doesn't requires a `StringComparison` for `Match` function. Three out of the five implementations of this interface didn't use this value.
2. The comparison type is decided when `PathSegment` are created instead of at execution time. Benefit of doing this is to simplify the code: the comparison type doesn't need to be passed down from method to method.
3. `PathBuilder` is not a static class any more. Because it nows hold on the value of comparison type.
4. Default comparison type is case-insensitive.

Tests are updated accordingly and the new tests are added to cover both cases.

Related: https://github.com/aspnet/dnx/issues/1645

/review: @davidfowl @Tratcher @victorhurdugaci 
/cc: @muratg @lodejard @pranavkm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ nuget.exe
 *.sln.ide
 debugSettings.json
 project.lock.json
+/.vs/
--- a/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Framework.FileSystemGlobbing.Abstractions;
 using Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments;
+using Microsoft.Framework.FileSystemGlobbing.Util;
 
 namespace Microsoft.Framework.FileSystemGlobbing.Internal
 {
@@ -16,19 +17,28 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
         private readonly IList<IPatternContext> _excludePatternContexts;
         private readonly IList<string> _files;
 
+        private readonly HashSet<string> _declaredLiteralFolderSegmentInString;
         private readonly HashSet<LiteralPathSegment> _declaredLiteralFolderSegments = new HashSet<LiteralPathSegment>();
         private readonly HashSet<LiteralPathSegment> _declaredLiteralFileSegments = new HashSet<LiteralPathSegment>();
 
         private bool _declaredParentPathSegment;
         private bool _declaredWildcardPathSegment;
 
-        public MatcherContext(IEnumerable<IPattern> includePatterns, IEnumerable<IPattern> excludePatterns, DirectoryInfoBase directoryInfo)
+        private readonly StringComparison _comparisonType;
+
+        public MatcherContext(IEnumerable<IPattern> includePatterns,
+                              IEnumerable<IPattern> excludePatterns,
+                              DirectoryInfoBase directoryInfo,
+                              StringComparison comparison)
         {
             _root = directoryInfo;
             _files = new List<string>();
+            _comparisonType = comparison;
 
             _includePatternContexts = includePatterns.Select(pattern => pattern.CreatePatternContextForInclude()).ToList();
             _excludePatternContexts = excludePatterns.Select(pattern => pattern.CreatePatternContextForExclude()).ToList();
+
+            _declaredLiteralFolderSegmentInString = new HashSet<string>(StringComparisonHelper.GetStringComparer(comparison));
         }
 
         public PatternMatchingResult Execute()
@@ -53,12 +63,12 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
             }
             else
             {
-                foreach (var literal in _declaredLiteralFolderSegments)
+                var candidates = directory.EnumerateFileSystemInfos().OfType<DirectoryInfoBase>();
+                foreach (var candidate in candidates)
                 {
-                    var dirInfo = directory.GetDirectory(literal.Value);
-                    if (dirInfo != null)
+                    if (_declaredLiteralFolderSegmentInString.Contains(candidate.Name))
                     {
-                        entities.Add(dirInfo);
+                        entities.Add(candidate);
                     }
                 }
             }
@@ -132,6 +142,7 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
                 else
                 {
                     _declaredLiteralFolderSegments.Add(literalSegment);
+                    _declaredLiteralFolderSegmentInString.Add(literalSegment.Value);
                 }
             }
             else if (patternSegment is ParentPathSegment)
--- a/src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
@@ -2,12 +2,15 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Framework.FileSystemGlobbing.Util;
 
 namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
 {
     public class LiteralPathSegment : IPathSegment
     {
-        public LiteralPathSegment(string value)
+        private readonly StringComparison _comparisonType;
+
+        public LiteralPathSegment(string value, StringComparison comparisonType)
         {
             if (value == null)
             {
@@ -15,13 +18,15 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
             }
 
             Value = value;
+
+            _comparisonType = comparisonType;
         }
 
         public string Value { get; }
 
-        public bool Match(string value, StringComparison comparisonType)
+        public bool Match(string value)
         {
-            return string.Equals(Value, value, comparisonType);
+            return string.Equals(Value, value, _comparisonType);
         }
 
         public override bool Equals(object obj)
@@ -29,12 +34,13 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
             var other = obj as LiteralPathSegment;
 
             return other != null &&
-                string.Equals(other.Value, Value, StringComparison.Ordinal);
+                _comparisonType == other._comparisonType &&
+                string.Equals(other.Value, Value, _comparisonType);
         }
 
         public override int GetHashCode()
         {
-            return Value.GetHashCode();
+            return StringComparisonHelper.GetStringComparer(_comparisonType).GetHashCode(Value);
         }
     }
 }
--- a/src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/WildcardPathSegment.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/WildcardPathSegment.cs
@@ -8,13 +8,19 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
 {
     public class WildcardPathSegment : IPathSegment
     {
-        public static readonly WildcardPathSegment MatchAll = new WildcardPathSegment(string.Empty, new List<string>(), string.Empty);
+        // It doesn't matter which StringComparison type is used in this MatchAll segment because 
+        // all comparing are skipped since there is no content in the segment.
+        public static readonly WildcardPathSegment MatchAll = new WildcardPathSegment(
+            string.Empty, new List<string>(), string.Empty, StringComparison.OrdinalIgnoreCase);
 
-        public WildcardPathSegment(string beginsWith, List<string> contains, string endsWith)
+        private readonly StringComparison _comparisonType;
+
+        public WildcardPathSegment(string beginsWith, List<string> contains, string endsWith, StringComparison comparisonType)
         {
             BeginsWith = beginsWith;
             Contains = contains;
             EndsWith = endsWith;
+            _comparisonType = comparisonType;
         }
 
         public string BeginsWith { get; }
@@ -23,7 +29,7 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
 
         public string EndsWith { get; }
 
-        public bool Match(string value, StringComparison comparisonType)
+        public bool Match(string value)
         {
             var wildcard = this;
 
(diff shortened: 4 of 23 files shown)


src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/WildcardPathSegment.cs
```diff
@@ -8,13 +8,17 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
 {
     public class WildcardPathSegment : IPathSegment
     {
-        public static readonly WildcardPathSegment MatchAll = new WildcardPathSegment(string.Empty, new List<string>(), string.Empty);
+        public static readonly WildcardPathSegment MatchAll = new WildcardPathSegment(
+            string.Empty, new List<string>(), string.Empty, StringComparison.OrdinalIgnoreCase);
```
troydai: The string comparison here doesn't matter. It is not used in actual comparing.

.gitignore
```diff
@@ -26,3 +26,5 @@ nuget.exe
 *.sln.ide
 debugSettings.json
 project.lock.json
+
```
Tratcher: blank line?

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
```diff
@@ -29,7 +33,7 @@ public override bool Equals(object obj)
             var other = obj as LiteralPathSegment;
 
             return other != null &&
-                string.Equals(other.Value, Value, StringComparison.Ordinal);
+                string.Equals(other.Value, Value, _comparisonType);
```
Tratcher: This pattern is pretty strange. I can now create two LiteralPathSegments with different StringComparisons which can cause a.Equals(b) != b.Equals(a).

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
```diff
@@ -29,7 +33,7 @@ public override bool Equals(object obj)
...
         }
 
         public override int GetHashCode()
```
Tratcher: GetHashCode should use `_comparisonType.GetHashCode(Value)`;

.gitignore
troydai: Will remove.

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
troydai: You're right. I'm wondering if it is even necessary to override the Equals here.

victorhurdugaci: You can compare the comparison type too :)

troydai: Yup. Good point!

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
troydai: Will update.

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
```diff
@@ -7,7 +7,9 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
 {
     public class LiteralPathSegment : IPathSegment
     {
-        public LiteralPathSegment(string value)
+        private readonly StringComparison _comparisonType;
+
+        public LiteralPathSegment(string value, StringComparison comparisonType)
```
pranavkm: We should probably be flowing a comparer around - it'll let the caller change the way comparisons are applied.

src/Microsoft.Framework.FileSystemGlobbing/Internal/Patterns/PatternBuilder.cs
```diff
@@ -8,12 +8,24 @@
...
         private static readonly char[] _slashes = new[] { '/', '\\' };
         private static readonly char[] _star = new[] { '*' };
 
-        public static IPattern Build(string pattern)
+        public PatternBuilder()
+        {
+            ComparisonType = StringComparison.OrdinalIgnoreCase;
+        }
+
+        public PatternBuilder(StringComparison comparisonType)
```
pranavkm: Doesn't look like these properties are used outside of `Build`. Why make the change to non-static when the parameter could be passed in?

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/WildcardPathSegment.cs
pranavkm: Can you add a comment to that effect?

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
pranavkm: Would it suffice to simply `Debug.Assert` in that case? It would be an odd case for the two comparisons to be out of sync, wouldn't it?

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
troydai: It makes all the methods signature very verbose, if you track the flow of the StringComparison.

test/Microsoft.Framework.FileSystemGlobbing.Tests/FunctionalTests.cs
```diff
@@ -76,7 +106,37 @@ public void IncludeCaseSensitive(string root, string includePattern, string[] ex
...
+            var matcher = new Matcher(StringComparison.Ordinal);
+            matcher.AddInclude("**/*.*");
+            matcher.AddExclude(excludePattern);
+
+            ExecuteAndVerify(matcher, root, expectedFiles.Select(f => root + "/" + f).ToArray());
+        }
+
+        [Theory]
+        [InlineData("src/project/compiler/preprocess/", "source.cs", new string[] { "preprocess-source1.cs",
+                                                                                    "sub/preprocess-source2.cs",
```
pranavkm: These indentations will be a pain to maintain if any of the previous string lengths change. Can we just use VS default indents here? Using a `MemberData` might pretty it up.

src/Microsoft.Framework.FileSystemGlobbing/Internal/Patterns/PatternBuilder.cs
troydai: Because it's cleaner. As I choose not to flow the StringComparison during the execution but set up at construction it make sense to ensure all the building component are consistent.

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
pranavkm: `StringComparison` -> `StringComparer`

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
```diff
@@ -29,12 +33,31 @@ public override bool Equals(object obj)
...
+                case StringComparison.OrdinalIgnoreCase:
+                    return StringComparer.OrdinalIgnoreCase.GetHashCode(Value);
+#if !DNXCORE50
+                case StringComparison.InvariantCulture:
+                    return StringComparer.InvariantCulture.GetHashCode(Value);
+                case StringComparison.InvariantCultureIgnoreCase:
+                    return StringComparer.InvariantCultureIgnoreCase.GetHashCode(Value);
+#endif
+                default:
+                    throw new InvalidOperationException($"Unexpected StringComparison type: {_comparisonType}");
```
troydai: I don't like this giant switch statement, but it's the so far the only way I find to get a `StringComparer` from a `StringComparison`. We need the former instance to get the hash code properly.

Another solution suggested by @Tratcher is to use `StringComparer` in other places. However it doesn't fly, since string compare methods like `string.BeginWiths` accepts `StringComparison` only. I didn't find a easy way to get `StringComparison` from `StringComparer` either.

If you have a better solution, please let me know.

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/WildcardPathSegment.cs
troydai: Will do.

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
troydai: Need `StringComparison` here: https://github.com/aspnet/FileSystem/blob/trdai/dnx1645/src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/WildcardPathSegment.cs#L39

Know a way to get `StringComparison` from `StringComparer` easily?

troydai:
Ping @davidfowl @lodejard @pranavkm @Tratcher @victorhurdugaci

Tratcher:
Still weird, but better.

troydai:
Ping @Tratcher @davidfowl @pranavkm 
Any further comments?

src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs
```diff
@@ -12,16 +13,27 @@ public class Matcher
     {
         private IList<IPattern> _includePatterns = new List<IPattern>();
         private IList<IPattern> _excludePatterns = new List<IPattern>();
+        private PatternBuilder _builder;
```
pranavkm: readonly

test/Microsoft.Framework.FileSystemGlobbing.Tests/FunctionalTests.cs
pranavkm: :(

pranavkm:
:shipit: from me. Have either @davidfowl or @Tratcher also sign off

troydai:
I found a bigger issue here.  More work is needed to enable this change on Linux. The key issue is that `DirectoryInfo.GetDirectories` of Mono is case sensitive. To enable case insensitive we need to fallback to get all the child filesystem object from current directory and then filter. The logic will also invalidate some of the original design because now it is possible that one literal string can match multiple results under case-insensitive policy.

Clearly the scope of the change is bigger.

troydai:
Ping @pranavkm @Tratcher @davidfowl @lodejard please help to review the last commit. It addresses the issue that on Mono `GetDirectory` is case sensitive.

troydai:
Ping

troydai:
Ping

Tratcher:
Seems ok. How impactful is the change to consuming code? Will you send separate PRs for them or just hot patch them?

troydai:
The impact to consuming code is minimal. The exposed interface `Matcher` is not changed. Unless there is user out there implement the _internal_ interfaces, no should would be affected.

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
```diff
@@ -53,13 +63,13 @@ private void Match(DirectoryInfoBase directory, string parentRelativePath)
             }
             else
             {
-                foreach (var literal in _declaredLiteralFolderSegments)
+                var candidates = directory.EnumerateFileSystemInfos().Where(info => info is DirectoryInfoBase).Cast<DirectoryInfoBase>();
```
pranavkm: `OfType<DirectoryInfoBase>()`?

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
```diff
@@ -53,13 +63,13 @@ private void Match(DirectoryInfoBase directory, string parentRelativePath)
...
+                var comparer = StringComparisonHelper.GetStringComparer(_comparisonType);
+
+                foreach (var c in candidates)
```
pranavkm: `c` -> `candidate`

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
```diff
@@ -53,13 +63,13 @@ private void Match(DirectoryInfoBase directory, string parentRelativePath)
...
                 {
-                    var dirInfo = directory.GetDirectory(literal.Value);
-                    if (dirInfo != null)
-                    {
-                        entities.Add(dirInfo);
-                    }
+                    _declaredLiteralFolderSegmentInString.Contains(c.Name);
```
pranavkm: This line doesn't seem to do anything?

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
```diff
@@ -1,13 +1,16 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Framework.FileSystemGlobbing.Util;
```
pranavkm: `System` namespaces first.

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
pranavkm: Isn't this the same as `EnumerateDirectories`?

pranavkm:
:watch:

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/LiteralPathSegment.cs
troydai: will update.

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
troydai: The `directory` is of type `DirectoryInfoBase` which doesn't have `EnumerateDirectories` API. (It's not `DirectoryInfo`.

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
```diff
@@ -53,12 +63,14 @@ private void Match(DirectoryInfoBase directory, string parentRelativePath)
             }
             else
             {
-                foreach (var literal in _declaredLiteralFolderSegments)
+                var candidates = directory.EnumerateFileSystemInfos().Where(info => info is DirectoryInfoBase).Cast<DirectoryInfoBase>();
+                var comparer = StringComparisonHelper.GetStringComparer(_comparisonType);
```
BrennanConroy: unused

troydai:
Updated :watch:

troydai:
Ping @davidfowl @pranavkm @Tratcher

troydai:
The change integrated with DNX has been tested in Universe Repo

PR #120, opened 2015-08-20
teach Matcher to return Stems for matches

anurse:
This is a pre-req for work on aspnet/dnx#848

The "stem" of a match is the part of the path matched from the first wildcard on. Or, if no wildcards are present, the simple name of the file. For example:

| Pattern | Match | Stem |
| ---------| -------- | ------ |
| `tools/install.ps1` | `tools/install.ps1` | `install.ps1` |
| `tools/*.ps1` | `tools/install.ps1` | `install.ps1` |
| `tools/**/*.ps1` | `tools/sub1/sub2/sub3/install.ps1` | `sub/sub1/sub2/sub3/install.ps1` |
| `a/b/**/c/**/*.txt` | `a/b/x/y/c/z/d.txt` | `x/y/c/z/d.txt`|

/cc @lodejard @davidfowl @troydai

### DIFF
--- a/src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.Framework.FileSystemGlobbing
+{
+    public struct FilePatternMatch : IEquatable<FilePatternMatch>
+    {
+        public string Path { get; }
+        public string Stem { get; }
+
+        public FilePatternMatch(string path, string stem)
+        {
+            Path = path;
+            Stem = stem;
+        }
+
+        public bool Equals(FilePatternMatch other)
+        {
+            return string.Equals(other.Path, Path, StringComparison.OrdinalIgnoreCase) &&
+                string.Equals(other.Stem, Stem, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public override bool Equals(object obj)
+        {
+            return Equals((FilePatternMatch)obj);
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCodeCombiner.Start()
+                .Add(Path)
+                .Add(Stem)
+                .CombinedHash;
+        }
+    }
+}
--- a/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
+++ b/src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
         private readonly DirectoryInfoBase _root;
         private readonly IList<IPatternContext> _includePatternContexts;
         private readonly IList<IPatternContext> _excludePatternContexts;
-        private readonly IList<string> _files;
+        private readonly IList<FilePatternMatch> _files;
 
         private readonly HashSet<string> _declaredLiteralFolderSegmentInString;
         private readonly HashSet<LiteralPathSegment> _declaredLiteralFolderSegments = new HashSet<LiteralPathSegment>();
@@ -32,7 +32,7 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
                               StringComparison comparison)
         {
             _root = directoryInfo;
-            _files = new List<string>();
+            _files = new List<FilePatternMatch>();
             _comparisonType = comparison;
 
             _includePatternContexts = includePatterns.Select(pattern => pattern.CreatePatternContextForInclude()).ToList();
@@ -85,9 +85,12 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
                 var fileInfo = entity as FileInfoBase;
                 if (fileInfo != null)
                 {
-                    if (MatchPatternContexts(fileInfo, (pattern, file) => pattern.Test(file)))
+                    var result = MatchPatternContexts(fileInfo, (pattern, file) => pattern.Test(file));
+                    if (result.IsSuccessful)
                     {
-                        _files.Add(CombinePath(parentRelativePath, fileInfo.Name));
+                        _files.Add(new FilePatternMatch(
+                            path: CombinePath(parentRelativePath, fileInfo.Name),
+                            stem: result.Stem));
                     }
 
                     continue;
@@ -155,7 +158,7 @@ namespace Microsoft.Framework.FileSystemGlobbing.Internal
             }
         }
 
-        private string CombinePath(string left, string right)
+        internal static string CombinePath(string left, string right)
         {
             if (string.IsNullOrEmpty(left))
             {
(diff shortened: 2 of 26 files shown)


anurse:
Travis failure seems unrelated...

anurse:
As far as I can tell, globbing is only used by DNX right now, which uses a submodule (that I'll update when I fix aspnet/dnx#848), but I'll follow the build and chase down reactions that need to happen as a result of the change in interface (Match now returns a list of objects instead of strings).

src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
```
troydai: License ...

_Your neighbourhood license police_

src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
```diff
@@ -0,0 +1,37 @@
...
+        public FilePatternMatch(string path, string stem)
+        {
+            Path = path;
+            Stem = stem;
+        }
+
+        public bool Equals(FilePatternMatch other)
+        {
+            return other != null &&
+                string.Equals(other.Path, Path, StringComparison.OrdinalIgnoreCase) &&
```
troydai: In case we're going to switch between case sensitive and insensitive in the future, we better prepare for that.

src/Microsoft.Framework.FileSystemGlobbing/Internal/PatternContexts/PatternContextRagged.cs
```diff
@@ -58,6 +58,11 @@ public sealed override void PushDirectory(DirectoryInfoBase directory)
                 frame.BacktrackAvailable += 1;
             }
 
+            if(frame.InStem)
```
troydai: Format

src/Microsoft.Framework.FileSystemGlobbing/Internal/PatternContexts/PatternContextLinear.cs
```diff
@@ -30,6 +31,14 @@ public override void PushDirectory(DirectoryInfoBase directory)
             }
             else
             {
+                // Determine this frame's contribution to the stem (if any)
+                var segment = Pattern.Segments[Frame.SegmentIndex];
+                if (Frame.InStem || segment.ProducesStem)
+                {
+                    Frame.InStem = true;
+                    Frame.Stem = MatcherContext.CombinePath(Frame.Stem, directory.Name);
```
troydai: I kind of concerns how much string will be created during the process here.

src/Microsoft.Framework.FileSystemGlobbing/Internal/MatcherContext.cs
```diff
@@ -167,36 +170,43 @@ private string CombinePath(string left, string right)
             }
         }
 
+        // Used to adapt Test(DirectoryInfoBase) for the below overload
         private bool MatchPatternContexts<TFileInfoBase>(TFileInfoBase fileinfo, Func<IPatternContext, TFileInfoBase, bool> test)
         {
-            var found = false;
+            return MatchPatternContexts(fileinfo, (ctx, file) => test(ctx, file) ? PatternTestResult.Success(stem: string.Empty) : PatternTestResult.Failed).IsSuccessful;
```
troydai: This is really difficult to read.

troydai:
Functional looks fine to me. But I'm a bit concern about the performance. With the change even we turn off `stem` generation, during the globbing `PatternTestResult` is created instead of a boolean. It puts pressure on GC. We need to profile the scenario.

src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
anurse: guh... It's such a pain. I wonder if we should have some kind of bot that can auto-add it :)

anurse: If I ever write such a bot, I'm calling it "auto-troy"

src/Microsoft.Framework.FileSystemGlobbing/Internal/PatternContexts/PatternContextLinear.cs
anurse: I'm changing it to be a list of strings that will be joined together at the end. This, plus making the FilePatternMatch object a struct means the only additional allocation is the final stem string. I considered making that lazy, but to do that you'd have to hold all the individual stem items, which would basically take up the same space.

I'll do some before/after memory tracing to confirm.

src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
anurse: this method isn't used for any matching, so I don't think it needs to be subject to `PatternBuilder.ComparisonType`

anurse:
Initial performance results show no notable increase in memory usage. I ran the globbing with a selection of highly-recursive globs over `C:\ProgramData\Microsoft` (which contains 48K files across 4K folders). The globs:

```
            matcher.AddInclude("**/1033/*");
            matcher.AddInclude("**/en-US/*");
            matcher.AddInclude("**/v4.0/*");
            matcher.AddInclude("**/*.dll");
            matcher.AddInclude("**/*.dat");
            matcher.AddInclude("**/*.tmp");
```

Some excludes were needed to avoid permission errors, they excluded a total of 11 files in 1 folder:

```
            matcher.AddExclude("Crypto/**");
            matcher.AddExclude("Diagnosis");
            matcher.AddExclude("Network/Downloader");
            matcher.AddExclude("Windows/SystemData");
```

Initial results are below (left is before, right is after):

image

The result is a change of 4MB, but other runs showed a decrease in memory:

image

Again, left is before change, right is after. Compare Snapshot 2 on left to Snapshot 3 on right, both taken at the same point in the test program.

Deeper analysis reveals that File System globbing increased in retained memory by appoximately 34KB and unretained? memory by ~50KB (see below, left before, right after). Continuing to dig deeper I discovered that this was almost entirely the final, composed, stems. I'll do some initial investigation in to allowing a caller to disable stem generation, but given that the Paths are already retained and the Stem can be no longer than the Path this doesn't seem like a massive impact.

image

ping @davidfowl

anurse:
:watch: @troydai @davidfowl

anurse:
pingity ping @davidfowl @troydai

src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public FilePatternMatch(string path, string stem)
+        {
+            Path = path;
+            Stem = stem;
+        }
+
+        public bool Equals(FilePatternMatch other)
+        {
+            return string.Equals(other.Path, Path, StringComparison.OrdinalIgnoreCase) &&
```
moozzyk: Should it be IgnoreCase? - paths are case sensitive on Linux/Mac

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/CurrentPathSegment.cs
```diff
@@ -1,10 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+
 namespace Microsoft.Framework.FileSystemGlobbing.Internal.PathSegments
 {
     public class CurrentPathSegment : IPathSegment
     {
+        public bool ProducesStem { get { return false; } }
```
moozzyk: Can we find a better name for this property (especially since it is public)? One that comes to my mind is `CanProduceStem`...

src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
anurse: correct, but globbing is case insensitive :)

anurse: see https://github.com/aspnet/FileSystem/blob/anurse/dnx-848-glob-stem/src/Microsoft.Framework.FileSystemGlobbing/Matcher.cs#L20

src/Microsoft.Framework.FileSystemGlobbing/Internal/PathSegments/CurrentPathSegment.cs
anurse: :+1:

src/Microsoft.Framework.FileSystemGlobbing/Internal/PatternContexts/PatternContextRaggedExclude.cs
```diff
@@ -13,19 +13,18 @@ public PatternContextRaggedExclude(IRaggedPattern pattern)
         {
         }
 
-        public override bool Test(FileInfoBase file)
+        public override PatternTestResult Test(FileInfoBase file)
```
moozzyk: Is there a good way to have just one copy of this method - the one PatternContextRaggedInclude.cs seems identical. Also do you mind improving the exception message?

src/Microsoft.Framework.FileSystemGlobbing/FilePatternMatch.cs
moozzyk: Thanks. I haven't really looked into globbing yet...

moozzyk:
LGTM

anurse:
Last call @davidfowl @lodejard

dazinator:
> correct, but globbing is case insensitive :)

Is it? I just read about globbing on wiki, but it seems to indicate that glob patterns are case sensitive, atleast on unix: https://en.wikipedia.org/wiki/Glob_(programming)

Perhaps you would want to make case sensitivity optional / configurable?

PR #133, opened 2015-10-14
EmbeddedFileProvider should allow paths with back slashes

### LINKED ISSUE #129: EmbeddedFileProvider cannot load embedded _ViewStart and _ViewImports
In GetFileInfo function, we replaced '/' with '.' to create resourcePath.

```
public IFileInfo GetFileInfo(string subpath)
{
    ...
    string resourcePath = _baseNamespace + subpath.Replace('/', '.');
    ...
}
```

However, when requesting _ViewStart and _ViewImports subpath comes with '\' instead of '/' causing resourcePath to be incorrect.

My suggestion is to replace both '/' and '\' with '.'.

pranavkm:
Fixes #129


pranavkm:
cc @Eilon

Eilon:
:shipit:

PR #136, opened 2015-10-22
Wait after directory creation to avoid race condition.

cesarbs:
See https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs#L13.

Some tests were very flaky on Linux because they acted on directories that had just been created. Introducing a delay after directory creation made the tests stable.

I did have to skip Token_Fired_On_Directory_Name_Change though, since it's flaky after the directory is renamed. Introducing a delay didn't help there.

### DIFF
--- a/test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
+++ b/test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
@@ -16,6 +16,9 @@ namespace Microsoft.AspNet.FileProviders
     {
         private const int WaitTimeForTokenToFire = 2 * 100;
 
+        // See https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs#L13
+        private const int WaitTimeAfterDirectoryCreation = 2 * 100;
+
         [Fact]
         public void ExistingFilesReturnTrue()
         {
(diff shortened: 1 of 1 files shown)


cesarbs:
@Tratcher @davidfowl

test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -16,6 +16,9 @@ public class PhysicalFileProviderTests
     {
         private const int WaitTimeForTokenToFire = 2 * 100;
 
+        // See https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem.Watcher/src/System/IO/FileSystemWatcher.Linux.cs#L13
```
Tratcher: Link to a specific commit or label or else this link will drift.

test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -669,7 +680,7 @@ public void Tokens_With_Forward_And_Backward_Slash()
         }
 
         [ConditionalFact]
-        [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
+        [OSSkipCondition(OperatingSystems.Linux, SkipReason = "Race condition in inotify API makes this very flaky.")]
```
Tratcher: Was there no reasonable amount of delay that would make this reliable? How long would you need?

cesarbs: I expirmented with different (reasonable) delays and none worked... The test fails 1 out of 5 times it's run.

Tratcher: Triage with @muratg to decide if this is an scenario we should address in production, or if the scenario is un-important and we should re-write the test.

cesarbs:
This will be resolved once https://github.com/aspnet/FileSystem/issues/140 is taken care of. Closing this PR.

PR #139, opened 2015-11-10
Use separate directory for each PhysicalFileProvider test

cesarbs:
Currently tests are using GetTempPath() as the root for file system watchers. Because those tests are run in parallel, this overwhelms the watchers, which causes them to raise error events because their filesystem event buffers get filled up. Tests that expected tokens not to be fired failed because of this, because our watcher fires tokens when there are errors in the CLR's watchers. Moreover, tests that expected tokens to be fire pass for the wrong reason i.e. the token is fired not because the test's condition has been satisfied but because there was an error in the associated FileSystemWatcher.

By using a separate temp directory for each test we ensure that no watchers are overwhelmed with events being triggered by other tests running concurrently.

cc @Tratcher @davidfowl @pranavkm

### DIFF
--- a/test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
+++ b/test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Testing.xunit;
+using Microsoft.Extensions.FileSystemGlobbing.Tests.TestUtility;
 using Microsoft.Extensions.Primitives;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


pranavkm:
Could we redesign these tests so that we're not testing the file watcher? The only thing we need to verify is if we trigger it for the right things (deletes \ adds \ modifies), not the FileSystemWatcher in itself.

pranavkm:
:watch:

cesarbs:
@pranavkm That's a good idea, but I'd rather address than on a future change. I really want to get these changes in asap so we can fix some instability in the CI.

pranavkm:
Cool. Looks :shipit: to me. Could you file a follow up bug to clean up these tests?

cesarbs:
@pranavkm https://github.com/aspnet/FileSystem/issues/140

Issue #49, opened 2015-01-21
Implement CombinedFileSystemProvider


glennc:
The CombinedFileSystemProvider would accept multiple FileSystemProviders and return files if they exist in one of them.

kichalla:
@muratg Fixing this issue can help in fixing https://github.com/aspnet/Diagnostics/issues/157 where razor files could be(as MVC gives this option) coming from physical or embedded resources

cc @davidfowl

kichalla:
Actually ignore my previous comment, the fix for aspnet/Diagnostics#157 is not dependent on fixing this issue...but from an end user perspective would be nice to have..

pranavkm:
Need this for https://github.com/aspnet/Mvc/issues/2551

glennc:
This is backlog for now. If it is definitely needed for MVC then come talk to us.

mgrosperrin:
Hi,
I'm really interested by this CombinedFileProvider.
Currently, I have my own implementation, but have to duplicate the not found file and directory (from the Microsoft.AspNet.FileProviders.Sources project).
Can I submit you a PR with this new file provider or is already someone working on it ?

pranavkm:
PR sounds great. I tried playing with this at one point and sort of dropped it because it was missing some design -

1. For GetFileInfo - does it keep querying until it finds a file?
2. For GetDirectoryContents - does it combine the results of multiple `IFileProvider` instances?

These might not be an issue if individual IFileProviders were isolated in separate hives (had to have distinct roots) but that didn't work for the scenarios we were trying to use it in Mvc.

mgrosperrin:
Hi,
OK for merging the content of multiple directories.
When I do this, should I consider that the Name property (of IFileInfo) is unique for a directory (ie when I have a file with a given name from a directory, I skip all files with the same name in the later directories)?

MartinJohns:
I was thinking a bit about this issue and here are my two cents:

1. `GetFileInfo`: This should iterate over all provided `IFileProvider` and return the first result where `Exists` is true. If there was no existing file found, then a `FileNotFound` class should be returned, as it is happening with all providers currently. This potentially leads to a lot of unnecessary garbage, because each provider will always instantiate a `FileNotFound` class which is thrown away immediately.

2. `GetDirectoryContents`: This should merge the contents of all provided `IFileProvider`. On duplicate entries (uniqueness on the `Name` property) it should take the first entry it found (again by the order of provided file providers).

[...]

PR #142, opened 2015-11-21
Implement a CombinedFileProvider

mgrosperrin:
This PR is for the issue #49 

The behavior for each method is:
1. `GetFileInfo`: returns the first `IFileInfo` of all provided `IFileProvider` where the `Exists` property is `true`,
2. `GetDirectoryContents`: merges the the contents of all provided `IFileProvider`. On duplicate entries (based on `Name` property), only the first  entry is kept,
3. `Watch`: returns a `IChangeToken` that merges the `IChangeToken` of all provided `IFileProvider`. When one of the token raises the change notification, the returned token raises it.


global.json
```diff
@@ -1,3 +1,6 @@
 {
-    "projects": ["src"]
+    "projects": [ "src" ],
+    "sdk": {
+        "version": "1.0.0-beta8"
```
MartinJohns: I don't think this modification should be part of the PR. Especially not since the latest stable version is RC1, not beta8.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,162 @@
...
+            // There is no existing directory contents
+            if (existingDirectoryContentsForAllProviders.Count == 0)
+            {
+                return new NotFoundDirectoryContents();
+            }
+            var combinedDirectoryContents = new CombinedDirectoryContents(existingDirectoryContentsForAllProviders);
+            return combinedDirectoryContents;
+        }
+
+        public IChangeToken Watch(string pattern)
```
MartinJohns: No documentation for this method? I think especially for this one it is interesting.

mgrosperrin:
Hi @MartinJohns,
thanks for your remarks. I've updated the global.json to remove my changes.
I have added documentation to the  `Watch` method, and add more info on the behavior of the others methods as well. I have put it in the `returns` part of the documentation. Let me know if this is good, or if it will be better in a `remarks` element.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        /// <summary>
+        /// Locates a file at the given path.
+        /// </summary>
+        /// <param name="subpath">The path that identifies the file. </param>
+        /// <returns>The file information. Caller must check Exists property. This will be the first existing <see cref="IFileInfo"/> returned by the provided <see cref="IFileProvider"/> or a not found <see cref="IFileInfo"/> if no existing files is found.</returns>
+        public IFileInfo GetFileInfo(string subpath)
+        {
+            foreach (var fileProvider in _fileProviders)
```
PinpointTownes: nit: you could change `_fileProviders`'s type to `IFileProvider[]` to avoid allocating an enumerator here (you don't have to explicitly use a `for` loop, since the compiler is supposed to optimize that for you).

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        /// <summary>
+        /// Enumerate a directory at the given path, if any.
+        /// </summary>
+        /// <param name="subpath">The path that identifies the directory</param>
+        /// <returns>Contents of the directory. Caller must check Exists property. The content is a merge of the contents of the provided <see cref="IFileProvider"/>. When there is multiple <see cref="IFileInfo"/> with the same Name property, only the first one is included on the results.</returns>
+        public IDirectoryContents GetDirectoryContents(string subpath)
+        {
+            // gets the content of all directories and merged them
+            var existingDirectoryContentsForAllProviders = _fileProviders.Select(
```
PinpointTownes: nit: for maximal performance, you could avoid using LINQ here.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            public bool Exists { get; }
+        }
+
+        private class CombinedFileChangeToken : IChangeToken
+        {
+            private readonly IEnumerable<IChangeToken> _changeTokens;
+
+            public CombinedFileChangeToken(IEnumerable<IChangeToken> changeTokens)
```
PinpointTownes: Same remark as above: since this parameter is always used with a list, you could replace IEnumerable with a list or an array.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                return NoopChangeToken.Singleton;
+            }
+            var combinedFileChangeToken = new CombinedFileChangeToken(activeChangeTokens);
+            return combinedFileChangeToken;
+        }
+
+        private class CombinedDirectoryContents : IDirectoryContents
+        {
+            private readonly Dictionary<string, IFileInfo> _files = new Dictionary<string, IFileInfo>();
```
PinpointTownes: Curious: why using a dictionary here? AFAICT, the key doesn't seem to be used anywhere.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            public bool ActiveChangeCallbacks
+            {
+                get { return _changeTokens.Any(_ => _.ActiveChangeCallbacks); }
+            }
+
+            private class Disposables : IDisposable
+            {
+                private readonly IEnumerable<IDisposable> _disposables;
```
PinpointTownes: nit: same remark :smile:

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+                }
+            }
+            return new NotFoundFileInfo(subpath);
+        }
+
+        /// <summary>
+        /// Enumerate a directory at the given path, if any.
+        /// </summary>
+        /// <param name="subpath">The path that identifies the directory</param>
+        /// <returns>Contents of the directory. Caller must check Exists property. The content is a merge of the contents of the provided <see cref="IFileProvider"/>. When there is multiple <see cref="IFileInfo"/> with the same Name property, only the first one is included on the results.</returns>
```
PinpointTownes: nit: this line seems a bit long :smile:

PinpointTownes:
Definitely a must have! :smile:

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,168 @@
...
+            return combinedFileChangeToken;
+        }
+
+        private class CombinedDirectoryContents : IDirectoryContents
+        {
+            private readonly Dictionary<string, IFileInfo> _files = new Dictionary<string, IFileInfo>();
+
+            public CombinedDirectoryContents(IEnumerable<IEnumerable<IFileInfo>> listOfFiles)
+            {
+                foreach (var files in listOfFiles)
```
PinpointTownes: This code path will immediately enumerate all the files exposed by the different providers, which may take some time. You should maybe consider deferring this operation to the last moment (in `GetEnumerator`) and using an iterator.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: I don't understand your point: `IFileProvider[]` implements `IEnumerable<IFileProvider>` so when an enumerator will be allocated (which will not be in your case) ?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: The dictionary is used to filter duplicate `IFileInfo` (see line 102). I can rewrite it to use a `Hash<string>` but I will have the hashset and the list of corresponding `IFileInfo`. I think it was simpler to do it this way but I can change.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: I have made the assumption that the directory contents will be pre-calculated (based on the two other providers), but you're right. I can move it to a `Lazy<IEnumerable<IFileInfo>>` (to avoiding dealing with the thread-safe and evaluating only once issues).

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: Yeah, it's the problem when VS is configured to do "Word wrap". I will add some new lines :smile:

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: OK, I will change it for a ~~old-school~~ traditionnal loop :cry:

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
PinpointTownes: When a variable is declared as an array, the compiler automatically replaces the `foreach` loop by a `for` loop and directly accesses the items via `array[index]` without having to instantiate a new enumerator. The compiler can't apply the same optimization when the type is `IEnumerable<>`.

mgrosperrin: Thanks, I'm happy to learn something today!

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+    /// </summary>
+    public class CombinedFileProvider : IFileProvider
+    {
+        private readonly IFileProvider[] _fileProviders;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CombinedFileProvider" /> class using a list of file provider.
+        /// </summary>
+        /// <param name="fileProviders"></param>
+        public CombinedFileProvider(params IFileProvider[] fileProviders)
```
davidfowl: I'd avoid params. Just take an `IEnumerable<IFileProvider>`

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// Enumerate a directory at the given path, if any.
+        /// </summary>
+        /// <param name="subpath">The path that identifies the directory</param>
+        /// <returns>Contents of the directory. Caller must check Exists property.
+        /// The content is a merge of the contents of the provided <see cref="IFileProvider"/>.
+        /// When there is multiple <see cref="IFileInfo"/> with the same Name property, only the first one is included on the results.</returns>
+        public IDirectoryContents GetDirectoryContents(string subpath)
+        {
+            // gets the content of all directories and merged them
+            var existingDirectoryContents = new List<IDirectoryContents>();
```
davidfowl: Instead of enumerating up front it might make sense to have the `CombinedDirectoryContents` take both file providers and do this in a more lazy matter (possible using yield or a custom iterator).

VaclavElias:
Just a question. Is this provider going to allow to have Views in Class Library? If yes, how can we use it? If not, please ignore my question :). Thanks.

MartinJohns:
@VaclavElias You can do that already.

```C#
services.Configure<RazorViewEngineOptions>(options =>
{
    options.FileProvider = yourFileProvider;
});
```

By the default the used file provider is a `PhysicalFileProvider , pointing to your views directory. You can replace this with a custom `IFileProvider`, e.g. the `EmbeddedFileProvider` to support views embedded in class libraries.

The here proposed `CombinedFileProvider` approach let's you use a `PhysicalFileProvider` as a first option, and the `EmbeddedFileProvider` as a second option. This would use views present in the local file system if they exist, with a fall back to the views embedded in the assembly.

VaclavElias:
Thanks @MartinJohns making it clear for me!

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
PinpointTownes: A `params` overload definitely makes sense, because that's how people will use this combined thingy: 

```csharp
options.FileProvider = new CombinedFileSystemProvider(
    new EmbeddedFileProvider(...),
    new EmbeddedFileProvider(...));
```

Since the file provider is a hot path, using `IEnumerable<IFileProvider>` is not really appropriate. Not to mention that, semantically, it's not immediately clear that the order matters when using `IEnumerable`. IMHO, an array or a list is much more appropriate in this case. That said, nothing prevents you from adding an `IEnumerable<>` overload that calls `ToArray()` or `ToList()` to avoid allocating enumerators in `GetFileInfo`/`GetDirectoryContents`.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
PinpointTownes: https://github.com/aspnet/FileSystem/pull/142#discussion_r45621235

That said, @mgrosperrin is right: the impact will be imperceptible when using `PhysicalFileProvider`, because it immediately stores the items returned by `EnumerateFileSystemInfos` in a new list when `GetDirectoryContents` is called.

That's definitely a surprising behavior, specially since it uses `EnumerableDirectoryContents`, that would allow using an iterator (or easier, a good `Select` LINQy thingy :smile:)

https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs#L146-L163

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+            public bool Exists { get { return true; } }
+        }
+
+        private class CombinedFileChangeToken : IChangeToken
+        {
+            private readonly List<IChangeToken> _changeTokens;
+
+            public CombinedFileChangeToken(List<IChangeToken> changeTokens)
+            {
+                _changeTokens = changeTokens ?? new List<IChangeToken>();
```
PinpointTownes: nit: looks like `changeTokens` cannot be null when used by `Watch`. I'd personally removed the null coalescing operator :smile:

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+            }
+
+            public IDisposable RegisterChangeCallback(Action<object> callback, object state)
+            {
+                var disposables = new List<IDisposable>();
+                foreach (var changeToken in _changeTokens)
```
PinpointTownes: Note: when using a list, the compiler doesn't apply any optimization. If you want to avoid the enumerator allocation here, you have to use a good old `for` loop.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+
+                public Disposables(List<IDisposable> disposables)
+                {
+                    _disposables = disposables;
+                }
+                public void Dispose()
+                {
+                    if (_disposables != null)
+                    {
+                        foreach (var disposable in _disposables)
```
PinpointTownes: Same remark :+1:

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+            // There is no change token with active change callbacks
+            if (activeChangeTokens.Count == 0)
+            {
+                return NoopChangeToken.Singleton;
+            }
+            var combinedFileChangeToken = new CombinedFileChangeToken(activeChangeTokens);
+            return combinedFileChangeToken;
+        }
+
+        private class CombinedDirectoryContents : IDirectoryContents
```
MartinJohns: I'm perhaps alone with this opinion and I've seen this in the other FileSystem projects already, but can't we stop with these nested classes and just make it an internal class within an own file?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
davidfowl: > A params overload definitely makes sense, because that's how people will use this combined thingy:

Not really. Just make an array if you need that:

```C#
new CombinedFileSystemProvider(new[] {
    new EmbeddedFileProvider(...),
    new EmbeddedFileProvider(...) });
```

I'd keep the params array overload for a static factory method instead of baking it into the ctor.


Converting it to an array or list is more appropriate than enforcing an array be passed in.

mgrosperrin: Hi @davidfowl, is it a general purpose to avoid params, or just in this situation ?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: OK for me.
I could submit a new PR in order to do it after this one to un-nest these classes in all projects.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+    {
+        private readonly IFileProvider[] _fileProviders;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CombinedFileProvider" /> class using a list of file provider.
+        /// </summary>
+        /// <param name="fileProviders"></param>
+        public CombinedFileProvider(params IFileProvider[] fileProviders)
+        {
+            _fileProviders = fileProviders;
```
pranavkm: `fileProviders` needs a null check.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+        /// <returns>An <see cref="IChangeToken"/> that is notified when a file matching <paramref name="filter"/> is added, modified or deleted.
+        /// The change token will be notified when one of the change token returned by the provided <see cref="IFileProvider"/> will be notified.</returns>
+        public IChangeToken Watch(string pattern)
+        {
+            // Watch all file providers
+            var activeChangeTokens = new List<IChangeToken>();
+            foreach (var fileProvider in _fileProviders)
+            {
+                var changeToken = fileProvider.Watch(pattern);
+                if (changeToken != null && changeToken.ActiveChangeCallbacks)
```
pranavkm: Why would it only look at `ActiveChangeCallbacks`?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
pranavkm: You could start with this - just make these types public.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        private class CombinedDirectoryContents : IDirectoryContents
+        {
+            private readonly Lazy<List<IDirectoryContents>> _directoriesContents;
+            private readonly Lazy<List<IFileInfo>> _files;
+            private readonly Lazy<bool> _exists;
+
+            public CombinedDirectoryContents(IFileProvider[] fileProviders, string subpath)
+            {
+                _directoriesContents =new Lazy<List<IDirectoryContents>>(() =>
```
pranavkm: Why make this deferred?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+            {
+                get { return _changeTokens.Any(_ => _.HasChanged); }
+            }
+
+            public bool ActiveChangeCallbacks
+            {
+                get { return _changeTokens.Any(_ => _.ActiveChangeCallbacks); }
+            }
+
+            private class Disposables : IDisposable
```
pranavkm: `CombinedDisposable` to maintain the naming pattern.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+                {
+                    var disposable = _changeTokens[i].RegisterChangeCallback(callback, state);
+                    disposables.Add(disposable);
+                }
+                return new Disposables(disposables);
+            }
+
+            public bool HasChanged
+            {
+                get { return _changeTokens.Any(_ => _.HasChanged); }
```
pranavkm: Don't use underscores for things you're using: 
`_changeTokens.Any(token => token.HasChanged)`. Same comment for `ActiveChangeCallbacks`

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+            {
+                private readonly List<IDisposable> _disposables;
+
+                public Disposables(List<IDisposable> disposables)
+                {
+                    _disposables = disposables;
+                }
+                public void Dispose()
+                {
+                    if (_disposables != null)
```
pranavkm: Isn't this guaranteed to be non null?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+using System.Linq;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.FileProviders
+{
+    /// <summary>
+    /// Looks up files using embedded resources in the specified assembly.
+    /// This file provider is case sensitive.
+    /// </summary>
+    public class CombinedFileProvider : IFileProvider
```
pranavkm: Should this be named `CompositeFileProvider`?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
```diff
@@ -0,0 +1,205 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.FileProviders
+{
+    /// <summary>
+    /// Looks up files using embedded resources in the specified assembly.
```
pranavkm: Summary is incorrrect.

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+
+namespace Microsoft.AspNet.FileProviders.Embedded.Tests
+{
+    public class CombinedFileProviderTests
+    {
+        [Fact]
+        public void GetFileInfo_ReturnsNotFoundFileInfo_IfNoFileProviderSpecified()
+        {
+            // Arrange
+            var provider = new CombinedFileProvider();
```
pranavkm: I would probably have it throw if there are fewer than 2 elements registered.

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+                    new MockFileInfo(fileName),
+                    new MockFileInfo("File3")
+                    )
+                );
+
+            // Act
+            var fileInfo = provider.GetFileInfo(fileName);
+
+            // Assert
+            Assert.NotNull(fileInfo);
```
pranavkm: You could replace the three statements with `Assert.Same(expectedFileInfo, fileInfo);`

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+            Assert.NotNull(files);
+            Assert.False(files.Exists);
+            Assert.Empty(files);
+        }
+
+        [Fact]
+        public void GetDirectoryContents_ReturnsCombinaisionOFFiles()
+        {
+            // Arrange
+            IFileInfo file1 = new MockFileInfo("File1"),
```
pranavkm: Use `var` and declare these as separate statements (instead of using commas)

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+                file2 = new MockFileInfo("File2"),
+                file2Bis = new MockFileInfo("File2"),
+                file3 = new MockFileInfo("File3");
+            var provider = new CombinedFileProvider(
+                new MockFileProvider(
+                    file1,
+                    file2
+                    ),
```
pranavkm: Formatting: put these parentheses on the previous line e.g. `file2),`

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+                );
+
+            // Act
+            var files = provider.GetDirectoryContents(string.Empty);
+
+            // Assert
+            Assert.NotNull(files);
+            Assert.True(files.Exists);
+            Assert.Collection(files.OrderBy(f => f.Name, StringComparer.Ordinal),
+                file => Assert.Equal(file1, file),
```
pranavkm: Use `Assert.Same` for reference equality.

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+                file => Assert.Equal(file2, file),
+                file => Assert.Equal(file3, file));
+        }
+
+        [Fact]
+        public void GetDirectoryContents_ReturnsCombinaisionOFFiles_WhenSomeFileProviderRetunsNoContent()
```
pranavkm: Typo: GetDirectoryContents_ReturnsCombinationOfFiles_WhenSomeFileProviderRetunsNoContent

src/Microsoft.AspNet.FileProviders.Combined/project.json
```diff
@@ -0,0 +1,29 @@
...
+      "type": "build"
+    },
+    "Microsoft.AspNet.FileProviders.Abstractions": "1.0.0-*"
+  },
+  "frameworks": {
+    "net451": {},
+    "dotnet5.4": {
+      "dependencies": {
+        "System.Collections": "4.0.11-*",
+        "System.Reflection": "4.0.10-*",
```
pranavkm: You might not need these.

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Microsoft.Extensions.Primitives;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.FileProviders.Combined.Tests.TestUtility
+{
+    internal class MockChangeToken : IChangeToken
```
pranavkm: Could you rename this to `TestChangeToken`? `MockChangeToken` sounds like it's using `Moq`

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Microsoft.Extensions.Primitives;
```
pranavkm: Could you add the license header to the file. Also sort usings so that `System` namespaces appear first.

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.AspNet.FileProviders.Combined.Tests.TestUtility;
+using Xunit;
+using System.Collections.Generic;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.FileProviders.Embedded.Tests
```
pranavkm: `namespace Microsoft.AspNet.FileProviders.Combined`

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.FileProviders.Combined.Tests.TestUtility
```
pranavkm: Could you move this file to the root of the test directory?

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
```diff
@@ -0,0 +1,57 @@
...
+                return true;
+            }
+        }
+
+        public bool HasChanged
+        {
+            get; set;
+        }
+
+        public int NbOfCallback
```
pranavkm: `public int CallbacksCount => _callbacks.Count`. Alternatively just expose the backing list as a property.

test/Microsoft.AspNet.FileProviders.Combined.Tests/project.json
```diff
@@ -0,0 +1,20 @@
...
+    "dependencies": {
+        "Microsoft.AspNet.FileProviders.Combined": "1.0.0-*",
+        "Microsoft.AspNet.FileProviders.Sources": "1.0.0-*",
+        "Microsoft.AspNet.Testing": "1.0.0-*",
+        "xunit.runner.aspnet": "2.0.0-aspnet-*"
+    },
+    "commands": {
+        "test": "xunit.runner.aspnet"
+    },
+    "resource": "File.txt;sub/**;Resources/**",
```
pranavkm: Shouldn't need this.

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockFileProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public IFileInfo GetFileInfo(string subpath)
+        {
+            var file = _files.FirstOrDefault(f => f.Name == subpath);
+            return file ?? new NotFoundFileInfo(subpath);
+        }
+
+        public IChangeToken Watch(string filter)
+        {
+            if(_changeTokens != null && _changeTokens.ContainsKey(filter))
```
pranavkm: I think it would be fine to null ref if we called watch without registering any changeTokens.

```C#
if (_changeTokens.TryGetValue(filter, out changeToken))
{
   return changeToken;
}
```

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockFileProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        {}
+
+        public MockFileProvider(params IFileInfo[] files)
+        {
+            _files = files;
+        }
+
+        public MockFileProvider(params KeyValuePair<string, IChangeToken>[] changeTokens)
+        {
+            _changeTokens = changeTokens.ToDictionary(changeToken => changeToken.Key,
```
pranavkm: Could you use vs default indentation:
```
 _changeTokens = changeTokens.ToDictionary(
      changeToken => changeToken.Key,
      changeToken => changeToken.Value,
      StringComparer.Ordinal);
```

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockFileProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNet.FileProviders.Combined.Tests.TestUtility
+{
+    internal class MockFileProvider : IFileProvider
```
pranavkm: public

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+
+        [Fact]
+        public void Watch_CombinedChangeToken()
+        {
+            // Arrange
+            var firstChangeToken = new MockChangeToken();
+            var secondChangeToken = new MockChangeToken();
+            var thirdChangeToken = new MockChangeToken();
+            var provider = new CombinedFileProvider(
+                new MockFileProvider(new KeyValuePair<string, IChangeToken>("pattern", firstChangeToken),
```
pranavkm: Use VS default indentation : https://github.com/aspnet/FileSystem/pull/142/files#r45886930

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,252 @@
...
+            Assert.Equal(0, firstChangeToken.NbOfCallback);
+            Assert.Equal(0, secondChangeToken.NbOfCallback);
+            Assert.Equal(0, thirdChangeToken.NbOfCallback);
+            var hasBeenCalled = false;
+            object result = null;
+            changeToken.RegisterChangeCallback(item =>
+            {
+                hasBeenCalled = true;
+                result = item;
+            }, null);
```
pranavkm: Could you add a test that makes sure the state flows to individual ChangeTokens?

pranavkm:
:watch:

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: This is the only way I found to determine if a `IChangeToken` is "real" and not a`NoopChangeToken`. I can combine all change tokens, but it will add some overload.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: This is in response to @davidfowl feedback: https://github.com/aspnet/FileSystem/pull/142#discussion_r45708539

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
mgrosperrin: I'm not sure that imposing a minimum number of providers is a good thing (it is not in one of my use case: I want the list to be configurable, and the configuration can be to take only one source).

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
mgrosperrin: I have named it as is to be consistent with the naming in the FileSystemGlobbing test project. But I can change it :smile:

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
mgrosperrin: As for the name, I have put it in this folder to be consistent with the FileSystemGlobbing test project.
Do I move the three files ?

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
mgrosperrin: Do I make the same change to all readonly properties (eg ActiveChangeCallbacks) ?

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
mgrosperrin: In fact, I should not combined `NoopChangeToken` as its `RegisterChangeCallback` method always raises a `NotSupportedException`.

mgrosperrin:
Hi,
I hope I have answer to all feedback.
The remaining points are (if I do not forget):
* The name of the provider: @pranavkm suggest `CompositeFileProvider`.
* The usage of `params` in the constructor: personnaly, I prefer the API with the `params` keyword, but @davidfowl doesn't seems to like it
* The check on the number of providers: @pranavkm suggest to impose to have at least two providers, but as I said, one of my use case is to combined the provider based on some configuration and I want to allow only one provider
* Should the implementation of `IDirectoryContents` and `IChangeToken` be public or internal ?
* Should I rename the fake implementation in the tests be renamed to `TestXXX`instead of `MockXXX` as in the FileSystemGlobbing test project ?
* Do I move these types to the root of the project instead of the `TestUtility` folder as in the FileSystemGlobbing test project ?
Thanks.

test/Microsoft.AspNet.FileProviders.Combined.Tests/TestUtility/MockChangeToken.cs
pranavkm: Yup. It's got one test file, no point burying it one level deep.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileProvider.cs
pranavkm: This would miss any change tokens that passively expire. My guess is you might have to register it but only set up change callbacks on ones that support active callbacks.

src/Microsoft.AspNet.FileProviders.Combined/CombinedDirectoryContents.cs
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.FileProviders
+{
+    internal class CombinedDirectoryContents : IDirectoryContents
```
pranavkm: `public`

src/Microsoft.AspNet.FileProviders.Combined/CombinedDirectoryContents.cs
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.FileProviders
```
pranavkm: @Eilon should these go under `Microsoft.AspNet.FileProviders.Combined`? The other two file providers are also placed under this namespace.

src/Microsoft.AspNet.FileProviders.Combined/CombinedDirectoryContents.cs
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    var directoryContents = fileProvider.GetDirectoryContents(subpath);
+                    if (directoryContents != null && directoryContents.Exists)
+                    {
+                        directories.Add(directoryContents);
+                    }
+                }
+                return directories;
+            }
+                );
```
pranavkm: Formatting: needs to go on the previous line.

src/Microsoft.AspNet.FileProviders.Combined/CombinedDirectoryContents.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    internal class CombinedDirectoryContents : IDirectoryContents
+    {
+        private readonly Lazy<List<IDirectoryContents>> _directoriesContents;
+        private readonly Lazy<List<IFileInfo>> _files;
+        private readonly Lazy<bool> _exists;
+
+        public CombinedDirectoryContents(IFileProvider[] fileProviders, string subpath)
+        {
+            _directoriesContents =new Lazy<List<IDirectoryContents>>(() =>
```
pranavkm: Formatting (needs spaces after the assignment). Bunch of places that have the same issue.

pranavkm: Could we possibly just do this on a first chance basis rather than use `Lazy`?  i.e. in 

```C#
public IEnumerator<IFileInfo> GetEnumerator()
{
    EnsureInitialized();
    return _files;
}

public bool Exists
{
   get
   {
      EnsureInitialized();
      return _exists;
  }
}

private void EnsureInitialized()
{
    if (_directoryContents == null)
    {
           // Do the stuff from the constructor here
     }
}

```

We aren't dealing with concurrency here, so just deferring evaluation should suffice.

src/Microsoft.AspNet.FileProviders.Combined/CombinedDisposable.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    internal class CombinedDisposable : IDisposable
+    {
+        private readonly List<IDisposable> _disposables;
+
+        public CombinedDisposable(List<IDisposable> disposables)
+        {
+            _disposables = disposables;
+        }
+        public void Dispose()
```
pranavkm: Need a blank line prior to this.

src/Microsoft.AspNet.FileProviders.Combined/CombinedDisposable.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            for (int i = 0; i < _disposables.Count; i++)
```
pranavkm: `var`

pranavkm: Couple of other places with the same issue.

src/Microsoft.AspNet.FileProviders.Combined/CombinedFileChangeToken.cs
```diff
@@ -0,0 +1,42 @@
...
+            var disposables = new List<IDisposable>();
+            for (int i = 0; i < _changeTokens.Count; i++)
+            {
+                var disposable = _changeTokens[i].RegisterChangeCallback(callback, state);
+                disposables.Add(disposable);
+            }
+            return new CombinedDisposable(disposables);
+        }
+
+        public bool HasChanged
```
pranavkm: You could rewrite this as 
`public bool HasChanged => _changeTokens.Any(token => token.HasChanged)`
Just a suggestion since it's not part of our guideline

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
pranavkm: It just feels weird when the CombinedFileProvider doesn't actually combine anything. Plus it's more work to cover these edge cases in tests

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,250 @@
...
+        public void GetFileInfo_ReturnsTheFirstFoundFileInfo()
+        {
+            // Arrange
+            var fileName = "File1";
+            var expectedFileInfo = new MockFileInfo(fileName);
+            var provider = new CombinedFileProvider(
+                new MockFileProvider(
+                    new MockFileInfo("FileA"),
+                    new MockFileInfo("FileB")
+                    ),
```
pranavkm: Formatting. Parentheses on previous lines

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,250 @@
...
+                file => Assert.Same(file1, file),
+                file => Assert.Same(file2, file),
+                file => Assert.Same(file3, file));
+        }
+
+        [Fact]
+        public void GetDirectoryContents_ReturnsCombinaitionOFFiles_WhenSomeFileProviderRetunsNoContent()
+        {
+            // Arrange
+            IFileInfo folderAFile1 = new MockFileInfo("FolderA/File1"),
```
pranavkm: `var`s on separate statements.

test/Microsoft.AspNet.FileProviders.Combined.Tests/CombinedFileProviderTests.cs
```diff
@@ -0,0 +1,250 @@
...
+            secondChangeToken.HasChanged = true;
+            Assert.False(changeToken.HasChanged);
+            secondChangeToken.HasChanged = false;
+            // third change token
+            thirdChangeToken.HasChanged = true;
+            Assert.True(changeToken.HasChanged);
+            thirdChangeToken.HasChanged = false;
+
+            // Register callback
+            Assert.Equal(0, firstChangeToken.Callsback.Count);
```
pranavkm: Could you split these into two tests? One's dealing with `HasChanged` and the other with callbacks.

pranavkm:
:watch:

pranavkm:
cc @Eilon for `CompositeFileProvider` versus `CombinedFileProvider`. Mvc has the habit of using `Composite` so it sounded like it would be a familiar pattern for users.

Eilon:
Yeah `Composite` sounds better to me for this.

mgrosperrin:
Hi,
is there anything else I can do to respond to your remarks?

pranavkm:
Sorry, GitHub doesn't send an email when you push an update. A comment saying you had it updated would be great (for future use). I'll have a look in a bit.

mgrosperrin:
Oh OK, didn't know that :smile: 
I'll know for the next time.

muratg:
@pranavkm Could you merge once it's good?

@mgrosperrin Thanks for the contribution!

src/Microsoft.AspNet.FileProviders.Composite/CompositeDirectoryContents.cs
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.FileProviders
+{
+    internal class CompositeDirectoryContents : IDirectoryContents
```
pranavkm: Could you make these types public? No reason to have them be internal.

src/Microsoft.AspNet.FileProviders.Composite/CompositeDirectoryContents.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        private readonly IFileProvider[] _fileProviders;
+        private readonly string _subPath;
+        private List<IFileInfo> _files;
+        private bool _exists;
+        private List<IDirectoryContents> _directories;
+
+        public CompositeDirectoryContents(IFileProvider[] fileProviders, string subpath)
+        {
+            _fileProviders = fileProviders;
```
pranavkm: This would need a null check (`throw new ArgumentNullException`)

src/Microsoft.AspNet.FileProviders.Composite/CompositeDirectoryContents.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    internal class CompositeDirectoryContents : IDirectoryContents
+    {
+        private readonly IFileProvider[] _fileProviders;
+        private readonly string _subPath;
+        private List<IFileInfo> _files;
+        private bool _exists;
+        private List<IDirectoryContents> _directories;
+
+        public CompositeDirectoryContents(IFileProvider[] fileProviders, string subpath)
```
pranavkm: `IList<IFileProvider>`

src/Microsoft.AspNet.FileProviders.Composite/CompositeFileProvider.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </summary>
+    public class CompositeFileProvider : IFileProvider
+    {
+        private readonly IFileProvider[] _fileProviders;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CompositeFileProvider" /> class using a list of file provider.
+        /// </summary>
+        /// <param name="fileProviders"></param>
+        public CompositeFileProvider(params IFileProvider[] fileProviders)
```
pranavkm: Did we settle on using `params`? I thought @davidfowl wanted to use an `IEnumerable` here?

pranavkm:
Looks good. If you could get the last few changes.

MartinJohns:
This might be a crazy idea given the state of so much code in the repositories.. But can't we start to comment the code? Add some documentation?

mgrosperrin:
Hi,
I have made the changes.
For the constructor of `CompositeFileProvider`, I have added an overload with an `IEnumerable` and kept the current one.

pranavkm:
I'll get this merged in soon.

pranavkm:
Merged in 15c9f1d229b000d9b2472b1fcfd22120a116669e

pranavkm:
Thanks for the PR!

mgrosperrin:
You're welcome!
Thanks for your reviews.

Issue #132, opened 2015-10-13
`PhysicalFilesWatcher` and `PhysicalFileProvider` should implement `IDisposable`


dougbu:
`PhysicalFilesWatcher` holds system resources in its `FileSystemWatcher` field. But `PhysicalFilesWatcher` and the (`public`) `PhysicalFileProvider` class containing the watcher do not implement `IDisposable`. System resources are therefore not released until finalization.

muratg:
@dougbu Was this a CR find, or were you investigating some leak?

dougbu:
Found this when investigating file system limits hit (aka `FileSystemWatcher` leaks) doing functional testing w/ DNX Core on Linux. We realized our use of `TestServer` was a mess for other reasons and moved those tests to use an xUnit fixture. That minimized our `TestServer` count enough for testing in this environment. But not all `PhysicalFileProvider` instances are pinned by the server so customers could hit similar issues.

@Tratcher and I discussed the issue back then and I finally got around to filing it.

PR #146, opened 2015-11-25
Make PhysicalFile Watcher and Provider disposable

BrennanConroy:
#132

### DIFF
--- a/test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
+++ b/test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
@@ -20,14 +20,16 @@ namespace Microsoft.AspNet.FileProviders
         [Fact]
         public void ExistingFilesReturnTrue()
         {
-            var provider = new PhysicalFileProvider(Directory.GetCurrentDirectory());
-            var info = provider.GetFileInfo("File.txt");
-            Assert.NotNull(info);
-            Assert.True(info.Exists);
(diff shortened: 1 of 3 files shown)


BrennanConroy:
Also, removed PhysicalFilesWatcher from PhysicalFileInfo because it wasn't used anywhere and not accessible to anyone.

BrennanConroy:
@Tratcher @dougbu

test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -20,14 +20,16 @@ public class PhysicalFileProviderTests
...
         {
-            var provider = new PhysicalFileProvider(Directory.GetCurrentDirectory());
-            var info = provider.GetFileInfo("File.txt");
-            Assert.NotNull(info);
-            Assert.True(info.Exists);
-
-            info = provider.GetFileInfo("/File.txt");
-            Assert.NotNull(info);
-            Assert.True(info.Exists);
+            using (var provider = new PhysicalFileProvider(Directory.GetCurrentDirectory()))
```
dougbu: Seems like all the changes from here on are about reducing file system burden while testing i.e. releasing resources ASAP. If there's an exception to that, let me know.

dougbu:
:shipit:

Do we have `Hosting`, `Mvc` and `StaticFiles` follow-ups planned to add `using`s, now that it's useful?

BrennanConroy:
Not really, I took a quick look at places this is used and I can change a few tests to dispose but can't easily dispose the other ones.

Issue #137, opened 2015-10-29
NotSupportedException for URLs with special characters


Tratcher:
Moved from http://katanaproject.codeplex.com/workitem/430

The static file middleware raises an exception when the input URL contains a character that is valid for HTTP but not authorized on a local path. Example: '/test:character' (note the colon). 
```
      System.NotSupportedException: The given path's format is not supported.
         at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
         at Microsoft.AspNet.FileProviders.PhysicalFileProvider.GetFullPath(String path)
         at Microsoft.AspNet.FileProviders.PhysicalFileProvider.GetDirectoryContents(String subpath)
         at Microsoft.AspNet.StaticFiles.DefaultFilesMiddleware.Invoke(HttpContext context) in C:\git\Universe\StaticFiles\src\Microsoft.AspNet.StaticFiles\DefaultFilesMiddleware.cs:line 66
         at Microsoft.AspNet.Hosting.Internal.RequestServicesContainerMiddleware.<Invoke>d__3.MoveNext()
      --- End of stack trace from previous location where exception was thrown ---
         at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
         at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
         at Microsoft.AspNet.Hosting.Internal.HostingEngine.<>c__DisplayClass32_0.<<Start>b__0>d.MoveNext()
      --- End of stack trace from previous location where exception was thrown ---
         at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
         at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
         at Microsoft.AspNet.Server.WebListener.MessagePump.<ProcessRequestAsync>d__23.MoveNext()
```

PhysicalFileProvider.GetDirectoryContents should not throw in this scenario.

muratg:
@BrennanConroy Another repo :)

BrennanConroy:
158dc437a9db16405436f533f6bd428d6820bc9c

PR #147, opened 2015-11-25
Don't throw with : in path

BrennanConroy:
#137 
@Tratcher

### DIFF
--- a/src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs
+++ b/src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs
@@ -45,12 +45,20 @@ namespace Microsoft.AspNet.FileProviders
 
         private string GetFullPath(string path)
         {
-            var fullPath = Path.GetFullPath(Path.Combine(Root, path));
-            if (!IsUnderneathRoot(fullPath))
+            try
             {
-                return null;
+                var fullPath = Path.GetFullPath(Path.Combine(Root, path));
+                if (!IsUnderneathRoot(fullPath))
+                {
+                    return null;
+                }
+                return fullPath;
             }
-            return fullPath;
+            catch (NotSupportedException)
+            {
+                // path contains a colon (":") that is not part of a volume identifier (for example, "c:\").
+            }
+            return null;
         }
 
         private bool IsUnderneathRoot(string fullPath)


Tratcher:
How about https://github.com/aspnet/StaticFiles/issues/84

BrennanConroy:
What about it?

muratg:
@BrennanConroy Is it supposed to be fixed here too?

BrennanConroy:
```/%3C``` is a perfectly valid string to pass in and causes no issues.

**Comment by davidfowl** [comment]
Yes, we shouldn't conflate the 2 issues. This might be slightly related but it's not the same bug.

**Comment by davidfowl** [comment]
@BrennanConroy  Where is the test for this fix?

**Comment by brentnewbury** [comment]
Sorry, I don't think I was quite clear in aspnet/StaticFiles#84 that `/%3C` is url decoded to `<` which causes an exception to be raised from the static files middleware when it's trying to determine the content type and whether to try to serve the file. I've not tried it, but from looking at the code, the file provider would also have a problem if the static files middleware carried on and tried to call `IFilerProvider.GetFileInfo(string)`. 

My opinion (and the fix I have waiting to create a pull request for [if you guys think what I have is okay]), is that the static file middleware see's the invalid characters in the request and skips on it, leaving it up to the next middleware in the pipeline... most likely ending up in a 404 returned to the client, not a 500.

File: src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs [hunk 1]
```diff
@@ -45,12 +45,20 @@ public PhysicalFileProvider(string root)
...
+                if (!IsUnderneathRoot(fullPath))
+                {
+                    return null;
+                }
+                return fullPath;
+            }
+            catch (NotSupportedException)
             {
-                return null;
+                // path contains a colon (":") that is not part of a volume identifier (for example, "c:\").
```
davidfowl: We should avoid first chance exceptions like these in cases where we know it will throw.

Tratcher:
I think this and aspnet/StaticFiles#84 are the same class of failure and should be addressed in similar ways. E.g. something like this: https://github.com/brentnewbury/StaticFiles/commit/11e02d1a513f41a8dc704027dd6a50e25f64054e#diff-565cc1cae6ec9b8df50ec6fb281418b4R120 rather than a try/catch.

BrennanConroy:
That looks good, but we'll still need to handle having a colon in the string

Tratcher:
@BrennanConroy Revert this.
/cc: @muratg

Tratcher:
You need to go through and make sure every call to a Path API with external data (e.g. a request path, but not the constructor) is guarded. Also, don't use try/catch, check for invalid characters explicitly.

davidfowl:
+1

Issue #104, opened 2015-04-26
Decide on handling path starting with / on *nix


davidfowl:
It's a broken assumption since absolute paths on *nix can start with `/`

Eilon:
@davidfowl we talked in theory about swapping these two checks: https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs#L85-L94

However, I don't think that's right either. The real problem is that on Windows it's easy to tell the difference between a volume-rooted path (`C:\...`) and an absolute path (`\foo\bar`), and on *nix there's no way to differentiate.

Honestly not quite sure how we could fix this even in theory...

davidfowl:
I think we have to drop this behavior completely.

cesarbs:
This somehow. affects https://github.com/aspnet/Universe/issues/304. I had to disable a unit test on *nix which ideally should be run on that platform. However the test gets incorrect results because of the way "/" is being treated.

muratg:
Let's reach a decision here. @davidfowl do you suggest we remove support for relative paths? 

Can we just enforce a `.` in front of relative paths?  (i.e. `./foo/bar/` instead of `/foo/bar`) to differentiate it from absolute paths?

davidfowl:
> @davidfowl do you suggest we remove support for relative paths?

No.

> Can we just enforce a . in front of relative paths? (i.e. `./foo/bar/` instead of `/foo/bar`) to differentiate it from absolute paths?

That's ugly. We can just not support paths that begin with /. Relative paths should just start with nothing. .e.g `foo/bar`.

muratg:
That sounds good to me.

muratg:
@pakrym this also prevents CoreCLR x-plat test runs.  Please look into this after your current item.

pranavkm:
> It's a broken assumption since absolute paths on *nix can start with /

Doesn't the `IFileProvider` root you at the rooted path? It's not meant to be a general purpose representation of the native file system.

davidfowl:
It's about what we allow in the methods. We want to fail for absolute paths.

muratg:
So do we fail on Windows for paths like C:\foo\bar as well?

davidfowl:
Yes I believe we already do

pakrym:
Decided that beheviour is okay, just changing supress message.
https://github.com/aspnet/FileSystem/pull/150

muratg:
[...]

PR #148, opened 2015-11-25
Remove support for relative path starting with /

pakrym:
https://github.com/aspnet/FileSystem/issues/104

@davidfowl @cesarbs

### DIFF
--- a/src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
+++ b/src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
@@ -63,16 +63,7 @@ namespace Microsoft.AspNet.FileProviders
 
             var builder = new StringBuilder(_baseNamespace.Length + subpath.Length);
             builder.Append(_baseNamespace);
-
-            // Relative paths starting with a leading slash okay
-            if (subpath.StartsWith("/", StringComparison.Ordinal))
-            {
-                builder.Append(subpath, 1, subpath.Length - 1);
-            }
-            else
-            {
-                builder.Append(subpath);
-            }
+            builder.Append(subpath);
 
             for (var i = _baseNamespace.Length; i < builder.Length; i++)
             {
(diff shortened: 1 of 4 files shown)


pranavkm:
You might also want to try this with Mvc. We pass paths starting with / all the time to the file system.

pranavkm:
Might also want to do something similar with `EmbeddedFileProvider`

davidfowl:
This also requires an announcement

test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -25,7 +25,7 @@ public void ExistingFilesReturnTrue()
             Assert.NotNull(info);
             Assert.True(info.Exists);
 
-            info = provider.GetFileInfo("/File.txt");
```
davidfowl: Need a negative test making sure that passing `/file` does the appropriate thing on each OS.

pakrym:
Made same changes to embedded file system and added negative tests

muratg:
@pakrym What's the behavior for Windows absolute paths such as `c:\foo\bar`?

pakrym:
We returning NoopFileInfo

src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -105,12 +96,6 @@ public IDirectoryContents GetDirectoryContents(string subpath)
...
             }
 
-            // Relative paths starting with a leading slash okay
-            if (subpath.StartsWith("/", StringComparison.Ordinal))
-            {
-                subpath = subpath.Substring(1);
-            }
-
             // Non-hierarchal.
             if (!subpath.Equals(string.Empty))
```
pranavkm: Could you combine this with the condition at Line 94?

test/Microsoft.AspNet.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -705,20 +722,16 @@ public void Tokens_With_Forward_And_Backward_Slash()
             using (var root = new DisposableFileSystem())
             {
                 var provider = new PhysicalFileProvider(root.RootPath);
-                var token1 = provider.Watch("/a/b");
-                var token2 = provider.Watch("a/b");
-                var token3 = provider.Watch(@"a\b");
+                var token1 = provider.Watch("a/b");
```
pranavkm: Could you make this into a `Theory`?

davidfowl:
We decided to punt this and deal with the inconsistency.

PR #151, opened 2015-12-08
Do not throw with invalid paths

BrennanConroy:
#137 
@Tratcher

### DIFF
--- a/src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
+++ b/src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 using Microsoft.Extensions.Primitives;
@@ -16,6 +17,8 @@ namespace Microsoft.AspNet.FileProviders
     /// </summary>
     public class EmbeddedFileProvider : IFileProvider
     {
+        private static readonly char[] _invalidFileNameChars = Path.GetInvalidFileNameChars()
+            .Where(c => c != Path.DirectorySeparatorChar && c != Path.AltDirectorySeparatorChar).ToArray();
         private readonly Assembly _assembly;
         private readonly string _baseNamespace;
         private readonly DateTimeOffset _lastModified;
@@ -83,6 +86,11 @@ namespace Microsoft.AspNet.FileProviders
             }
 
             var resourcePath = builder.ToString();
+            if (HasInvalidPathChars(resourcePath))
+            {
+                return new NotFoundFileInfo(resourcePath);
+            }
+
             var name = Path.GetFileName(subpath);
             if (_assembly.GetManifestResourceInfo(resourcePath) == null)
             {
@@ -140,6 +148,11 @@ namespace Microsoft.AspNet.FileProviders
         public IChangeToken Watch(string pattern)
         {
             return NoopChangeToken.Singleton;
+        }
+
+        private static bool HasInvalidPathChars(string path)
+        {
+            return path.IndexOfAny(_invalidFileNameChars) != -1;
         }
 
         private class EmbeddedResourceFileInfo : IFileInfo
--- a/src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs
+++ b/src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs
@@ -14,6 +14,8 @@ namespace Microsoft.AspNet.FileProviders
     /// </summary>
     public class PhysicalFileProvider : IFileProvider, IDisposable
     {
+        private static readonly char [] _invalidFileNameChars = Path.GetInvalidFileNameChars()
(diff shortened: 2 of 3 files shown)


Tratcher:
Does EmbeddedFileProvider have any similar issues?

src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -92,6 +97,11 @@ public IFileInfo GetFileInfo(string subpath)
                 subpath = subpath.Substring(1);
             }
 
+            if (HasInvalidPathChars(subpath))
```
Tratcher: You can combine this with the IsNullOrEmpty check.

src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -133,6 +143,11 @@ public IDirectoryContents GetDirectoryContents(string subpath)
                     subpath = subpath.Substring(1);
                 }
 
+                if (HasInvalidPathChars(subpath))
```
Tratcher: Move up

Tratcher:
@blowdart For https://github.com/aspnet/StaticFiles/issues/90 should we have HasInvalidPathChars also block `\` on Windows?

blowdart:
If that comes from a URL yes

src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -100,7 +100,7 @@ public IFileInfo GetFileInfo(string subpath)
         public IDirectoryContents GetDirectoryContents(string subpath)
         {
             // The file name is assumed to be the remainder of the resource name.
-            if (subpath == null)
+            if (subpath == null || HasInvalidPathChars(subpath))
```
pranavkm: This doesn't allow slashes in paths (other than a leading one). You could remove this check.

BrennanConroy: Yes it does, look at the HasInvalidPathChars function

src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -142,6 +142,15 @@ public IChangeToken Watch(string pattern)
             return NoopChangeToken.Singleton;
         }
 
+        private static bool HasInvalidPathChars(string path)
+        {
+            // convert path separators to a safe character so we can check for invalid file name characters
+            // and not get false positives from '/' or '\\'
+            var subPath = path.Replace(Path.DirectorySeparatorChar, '.')
```
pranavkm: Could we rewrite this method to avoid performing these `Replace`s? Just avoiding the extra allocations.

src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
pranavkm: See https://github.com/aspnet/FileSystem/pull/151/files#diff-2ca2083af78587a3714f664b5bddcb44R115

pranavkm: To clarify - this is `EmbeddedFileProvider` only allows listing `/` or empty strings.

BrennanConroy:
:arrow_up: :date:

src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -142,6 +147,11 @@ public IChangeToken Watch(string pattern)
             return NoopChangeToken.Singleton;
         }
 
+        private static bool HasInvalidPathChars(string path)
+        {
+            return path.IndexOfAny(Path.GetInvalidFileNameChars()) != -1;
```
Tratcher: Cache GetInvalidFileNameChars

src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -83,6 +86,11 @@ public IFileInfo GetFileInfo(string subpath)
             }
 
             var resourcePath = builder.ToString();
+            if(HasInvalidPathChars(resourcePath))
```
pranavkm: Ctrl K - D

src/Microsoft.AspNet.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -19,6 +20,8 @@ public class EmbeddedFileProvider : IFileProvider
         private readonly Assembly _assembly;
         private readonly string _baseNamespace;
         private readonly DateTimeOffset _lastModified;
+        private static readonly char[] _invalidFileNameChars = Path.GetInvalidFileNameChars()
```
pranavkm: Being pedantic, but `static readonly` before non-statics.

pranavkm:
:shipit:

Issue #144, opened 2015-11-24
Matcher returns wrong stems


reliak:
This addresses issue https://github.com/aspnet/Announcements/issues/58

I have the following folder structure:
- root
  - subfolder
    - sub1
    - sub2
    - sub3

I wanna find all files via:
```
var m = new Matcher();
m.AddInclude(@"**/*.*");
var res = m.Execute(new DirectoryInfoWrapper(new DirectoryInfo(root)));
```
As shown below, the Stems for files in `sub2` and `sub3` are wrong.
![globissue](https://cloud.githubusercontent.com/assets/4062774/11366023/d827b936-92aa-11e5-8dec-f0d500d18d79.png)

When I use a glob pattern starting with `subfolder` (see below) instead of the wildcards, all stems are correct:
```
var m = new Matcher();
m.AddInclude(@"subfolder/**/*.*");
var res = m.Execute(new DirectoryInfoWrapper(new DirectoryInfo(root)));
```
**Update**: Apparently the problem is not the glob pattern at the beginning. I inserted a folder under `subfolder`. When I now use the above pattern (`"subfolder/**/*.*"`) I also get wrong stems.

/cc @anurse

anurse:
I think I've seen this before. @muratg we should work on this for RC2.

muratg:
Agreed!

BrennanConroy:
a6ffb85a267cdc8f5985363bfc909904650003d5

PR #152, opened 2015-12-09
Return correct stems from match

BrennanConroy:
#144 

@anurse

### DIFF
--- a/test/Microsoft.Extensions.FileSystemGlobbing.Tests/FunctionalTests.cs
+++ b/test/Microsoft.Extensions.FileSystemGlobbing.Tests/FunctionalTests.cs
@@ -338,6 +338,62 @@ namespace Microsoft.Extensions.FileSystemGlobbing.Tests
                 "src/project2/source1.cs");
         }
 
+        [Fact]
+        public void StemCorrectWithDifferentWildCards()
+        {
+            var matcher = new Matcher();
+            matcher.AddInclude("sub/*.cs");
+            matcher.AddInclude("**/*.cs");
+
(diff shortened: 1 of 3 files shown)


test/Microsoft.Extensions.FileSystemGlobbing.Tests/FunctionalTests.cs
```diff
@@ -338,6 +338,37 @@ public void SingleFileAndRecursive()
                 "src/project2/source1.cs");
         }
 
+        [Fact]
+        public void StemCorrectWithDifferentWildCards()
```
anurse: Was this the only case that was broken before?

BrennanConroy: Any match that had multiple sub directories under it would produce incorrect stems for the files in the sub directories

davidfowl: Does this fix https://github.com/aspnet/dnx/issues/2750

anurse: Can you add a more specific test case for that as well then?

BrennanConroy: @davidfowl Yes, just tested it
@anurse Ok

BrennanConroy:
:up: :date:

anurse:
:shipit:

Issue #140, opened 2015-11-11
Refactor PhysicalFileProvider unit tests


cesarbs:
See @pranavkm's comment here: https://github.com/aspnet/FileSystem/pull/139#issuecomment-155584469

PR #153, opened 2015-12-10
Don't test FileSystemWatcher and FileInfo in PhysicalFileProvider tests (#140)

cesarbs:
1. Don't rely on FileSystemWatcher behavior for tests.
2. Add a few new tests.

https://github.com/aspnet/FileSystem/issues/140

cc @pranavkm @Tratcher @BrennanConroy


src/Microsoft.AspNet.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -199,7 +203,7 @@ public IChangeToken Watch(string filter)
             return _filesWatcher.CreateFileChangeToken(filter);
         }
 
-        private class PhysicalFileInfo : IFileInfo
+        internal class PhysicalFileInfo : IFileInfo
```
pranavkm: Make these `public`. No reason for these types to be internal. Also move it to a separate file.

BrennanConroy:
Fix Travis, but otherwise looks good

cesarbs:
Fixed.

PR #163, opened 2016-02-01
Prevent navigation above the root of PhysicalFileProvider

cesarbs:
cc @blowdart

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Extensions.FileProviders
     {
         private static readonly char[] _invalidFileNameChars = Path.GetInvalidFileNameChars()
             .Where(c => c != Path.DirectorySeparatorChar && c != Path.AltDirectorySeparatorChar).ToArray();
+        private static readonly char[] _pathSeparators = new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
         private readonly PhysicalFilesWatcher _filesWatcher;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -57,14 +57,49 @@ public void Dispose()
...
             {
                 return null;
             }
+
             return fullPath;
         }
 
+        private bool PathNavigatesAboveRoot(string path)
+        {
+            var tokenizer = new StringTokenizer(path, new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar });
```
pranavkm: Cache the array?

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -57,14 +57,49 @@ public void Dispose()
 
         private string GetFullPath(string path)
         {
+            if (PathNavigatesAboveRoot(path))
```
pranavkm: Wouldn't `IsUnderneathRoot` take care of rooting this? If not, could you add a comment about scenarios this method prevents that `IsUnderneathRoot` does not?

Eilon: Talk to @blowdart directly. This is needed.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
cesarbs: :+1:

PR #168, opened 2016-03-01
Disable mono watcher

JunTaoLuo:
This fixes Travis which currently fails with the message `System.IO.IOException : kqueue() FileSystemWatcher has reached the maximum nunmber of files to watch.` (https://travis-ci.org/aspnet/FileSystem/jobs/112436019)


davidfowl:
:sheep: :it:

PR #169, opened 2016-03-03
Target minimal TFMs

pranavkm:


pranavkm:
cc @davidfowl

src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.Extensions.FileProviders
     public class EmbeddedFileProvider : IFileProvider
     {
         private static readonly char[] _invalidFileNameChars = Path.GetInvalidFileNameChars()
-            .Where(c => c != Path.DirectorySeparatorChar && c != Path.AltDirectorySeparatorChar).ToArray();
+            .Where(c => c != '/' && c != '\\').ToArray();
```
pranavkm: Doing this lets us not cross-compile this project.

pranavkm:
Updated

src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
rynowak: But what if they change the path separator character to :trollface:

Eilon:
:shipit:

PR #173, opened 2016-03-11
Target dnx451

bricelam:
Reverts f966d4e5f722fff62bc5c2b2a1c931d03ca2a05a for DNX


bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

Issue #174, opened 2016-03-14
Add NullFileProvider


JunTaoLuo:
Need to port https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/NullFileProvider.cs back to FileSystem.

cc @davidfowl

PR #175, opened 2016-03-15
Add NullFileProvider

JunTaoLuo:
https://github.com/aspnet/FileSystem/issues/174

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider.cs
+++ b/src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders
+{
+    /// <summary>
+    /// An empty file provider with no contents.
+    /// </summary>
+    public class NullFileProvider : IFileProvider
+    {
+        /// <summary>
+        /// Enumerate a non-existent directory.
+        /// </summary>
+        /// <param name="subpath">A path under the root directory. This parameter is ignored.</param>
+        /// <returns>A <see cref="IDirectoryContents"/> that does not exist and does not contain any contents.</returns>
+        public IDirectoryContents GetDirectoryContents(string subpath) => new NotFoundDirectoryContents();
+
+        /// <summary>
+        /// Locate a non-existent file.
+        /// </summary>
+        /// <param name="subpath">A path under the root directory.</param>
+        /// <returns>A <see cref="IFileInfo"/> representing a non-existent file at the given path.</returns>
+        public IFileInfo GetFileInfo(string subpath) => new NotFoundFileInfo(subpath);
+
+        /// <summary>
+        /// Returns a <see cref="IChangeToken"/> that monitors nothing.
+        /// </summary>
+        /// <param name="filter">Filter string used to determine what files or folders to monitor. This parameter is ignored.</param>
+        /// <returns>A <see cref="IChangeToken"/> that does not register callbacks.</returns>
+        public IChangeToken Watch(string filter) => NoopChangeToken.Singleton;
+    }
+}
--- a/src/Microsoft.Extensions.FileProviders.Sources/project.json
+++ b/src/Microsoft.Extensions.FileProviders.Sources/project.json
@@ -10,9 +10,6 @@
     "url": "git://github.com/aspnet/filesystem"
   },
   "shared": "**/*.cs",
-  "dependencies": {
-    "Microsoft.Extensions.FileProviders.Abstractions": "1.0.0-*"
-  },
   "frameworks": {
     "netstandard1.0": {
       "dependencies": {
--- a/FileSystem.sln
+++ b/FileSystem.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+VisualStudioVersion = 14.0.24720.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{A1477614-E825-4204-A684-385004B63AEB}"
 EndProject
--- a/src/Microsoft.Extensions.FileProviders.Abstractions/project.json
+++ b/src/Microsoft.Extensions.FileProviders.Abstractions/project.json
@@ -18,6 +18,10 @@
     "url": "git://github.com/aspnet/filesystem"
   },
   "dependencies": {
+    "Microsoft.Extensions.FileProviders.Sources": {
+      "version": "1.0.0-*",
+      "type": "build"
+    },
     "Microsoft.Extensions.Primitives": "1.0.0-*"
   },
   "frameworks": {
--- a/src/Microsoft.Extensions.FileProviders.Sources/NotFoundDirectoryContents.cs
+++ b/src/Microsoft.Extensions.FileProviders.Sources/NotFoundDirectoryContents.cs
@@ -9,23 +9,10 @@ namespace Microsoft.Extensions.FileProviders
 {
     internal class NotFoundDirectoryContents : IDirectoryContents
     {
-        public NotFoundDirectoryContents()
-        {
-        }
+        public bool Exists => false;
 
-        public bool Exists
-        {
-            get { return false; }
-        }
+        public IEnumerator<IFileInfo> GetEnumerator() => Enumerable.Empty<IFileInfo>().GetEnumerator();
 
-        public IEnumerator<IFileInfo> GetEnumerator()
-        {
-            return Enumerable.Empty<IFileInfo>().GetEnumerator();
-        }
-
-        IEnumerator IEnumerable.GetEnumerator()
-        {
-            return GetEnumerator();
-        }
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     }
 }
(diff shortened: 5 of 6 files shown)


JunTaoLuo:
Hmm maybe I should add some tests.

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullFileInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public long Length => -1;
+
+        public string Name { get; }
+
+        public string PhysicalPath => null;
+
+        public Stream CreateReadStream()
+        {
+            throw new FileNotFoundException(string.Format($"{nameof(NullFileProvider)} does not support reading files.", Name));
```
MartinJohns: Why are you using string interpolation AND `string.Format` here? Also, you pass "Name" as the second argument to `string.Format`, but there is no `{0}`.

JunTaoLuo: Good point. In fact I think the code where I copied this from originally intended the `Name` to be the second argument of `FileNotFoundException()`

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullFileInfo.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
pranavkm: Can't you just use the `NotFoundFileInfo` instead?

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullFileProvider.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Abstractions
+{
+    public class NullFileProvider : IFileProvider
```
pranavkm: Should this go in `Microsoft.Extensions.FileProviders.Null` seeing that's how the rest of the types here are structured?

pranavkm:
:watch:

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullFileInfo.cs
JunTaoLuo: Actually I think we can also use `NoopChangeToken` and `NotFoundDirectoryContents` as well.

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullDirectoryContents.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.Extensions.FileProviders.Abstractions
+{
+    public class NullDirectoryContents : IDirectoryContents
+    {
+        public bool Exists => false;
+
+        public IEnumerator<IFileInfo> GetEnumerator()
+        {
+            yield break;
```
MartinJohns: Wouldn't it be better to use `Enumerable.Empty<IFileInfo>`? Much more declarative.

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullFileInfo.cs
JunTaoLuo: Though currently Sources depend on Abstractions so there would be a circular dependency there. I think we should just move the NullFileProvider to Sources as well since it is an implementation anyways.

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullDirectoryContents.cs
davidfowl: `public IEnumerator<IFileInfo> GetEnumerator() => Enumerable.Empty<IFileInfo>()`

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullFileInfo.cs
JunTaoLuo: @davidfowl I think the layering here is a little strange. Since most of the helper classes already exist in the `Sources` project which we do not publish but should reuse, we should just create another project that contains the NullFileProvider which reference Sources like the other FileProviders.

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider/NullDirectoryContents.cs
pranavkm: Or just use `NotFoundDirectoryContents`. None of these types are required.

JunTaoLuo: That's what I mentioned above. But we need to change the layering. We need to move NullFileProvider to its own project or else we'll get a circular dependency.

pranavkm: Yeah, that sounds fine. It'll be in line with `.Physical` \ `.Embedded`

JunTaoLuo:
:up::date:

src/Microsoft.Extensions.FileProviders.Null/project.json
```diff
@@ -0,0 +1,34 @@
...
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "warningsAsErrors": true,
+    "keyFile": "../../tools/Key.snk",
+    "nowarn": [
+      "CS1591"
+    ],
+    "xmlDoc": true
+  },
+  "description": "Null file and directory providers for Microsoft.Extensions.FileProviders.",
```
Eilon: This is not very descriptive...

Eilon:
Can we get a better name for this project? It looks like an error right now. Even something like NullProvider or something?

src/Microsoft.Extensions.FileProviders.Null/NullFileProvider.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Null
+{
+    public class NullFileProvider : IFileProvider
+    {
+        public IDirectoryContents GetDirectoryContents(string subpath)
```
pranavkm: Embrace your inner C#:
`public IDirectoryContents GetDirectoryContents(string subpath) => new NotFoundDirectoryContents();`

pranavkm:
:shipit: when @Eilon is happy.

src/Microsoft.Extensions.FileProviders.Null/project.json
davidfowl: This just needs to be a class in the abstractions project. It has no dependencies.

JunTaoLuo: That won't work if we want to reuse the types from `Microsoft.Extensions.FileProviders.Sources`. To avoid the circular dependency, we would either move those types to abstractions or move the null provider to sources both of which feels like incorrect layering.

davidfowl:
This shouldn't be a new project, just a class inside the abstractions

JunTaoLuo:
@Eilon See if you like the new description. Alternative package names could be `Placeholder`, `Noop`, `Emtpy`, `Dummy` and we can tack on a `Provider` at the end of those name too.

@davidfowl My comments keep on getting collapsed so I'll move it here:
>That won't work if we want to reuse the types from `Microsoft.Extensions.FileProviders.Sources`. To avoid the circular dependency, we would either move those types to abstractions or move the null provider to sources both of which feels like incorrect layering.

davidfowl:
Reusing helpers is unimportant. The layering is important. We shouldn't have a whole new project for the null provider.

JunTaoLuo:
Then I'll just move the other types to Abstractions then, there's no point duplicating them in sources.

Eilon:
Actually what's the issue in .Sources? The .Sources packages don't even compile so it doesn't matter what's in them.

JunTaoLuo:
Final decision is to keep the NullFileProvider inside `.Abstractions` and remove the incorrect dependency in `.Sources` instead.

JunTaoLuo:
:up::date:

src/Microsoft.Extensions.FileProviders.Sources/project.json
```diff
@@ -10,9 +10,7 @@
     "url": "git://github.com/aspnet/filesystem"
   },
   "shared": "**/*.cs",
-  "dependencies": {
-    "Microsoft.Extensions.FileProviders.Abstractions": "1.0.0-*"
-  },
+  "dependencies": {},
```
pranavkm: Remove?

pranavkm: Alternatively move the dependencies from the `frameworks` node here.

JunTaoLuo: Sure

src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders
+{
+    public class NullFileProvider : IFileProvider
```
pranavkm: Docs?

pranavkm:
:shipit: once you add the doc comments.

PR #176, opened 2016-03-21
Undo package changes

ryanbrandenburg:
CC @pranavkm


pranavkm:
:shipit:

MartinJohns:
It would be great to add some kind of description on **why** changes were made, so you can understand the reasoning behind the various changes in history.

PR #183, opened 2016-04-12
Add support for polling file watcher

### LINKED ISSUE #181: Augment FileChangeToken to use timestamps in addition to FileSystemWatcher's callbacks
Follow up to https://github.com/aspnet/FileSystem/pull/180. Our current mechanism of relying on `FileSystemWatcher` might have reliability issues with network shares. The plan is to additionally use a file system check to ensure the change token is eventually consistent. We'll turn this behavior on when an environment variable is set.

pranavkm:
Fixes #181


src/Microsoft.Extensions.FileProviders.Physical/PollingFileChangeToken.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        private DateTime GetLastWriteTimeUtc()
+        {
+            _fileInfo.Refresh();
+            return _fileInfo.Exists ? _fileInfo.LastWriteTimeUtc : DateTime.MinValue;
+        }
+
+        public bool ActiveChangeCallbacks => false;
+
+        public bool HasChanged
```
pranavkm: We might have to lock here to ensure that this behaves correctly when polled concurrently. A lot of these complexities show up because we have wildcard support in `Watch` and want to reuse instances of change token to avoid perform expensive regex matches. We could simplify a bunch of this here if we only ever support non-wildcard patterns.

cc @muratg \ @davidfowl thoughts?

muratg: I think it's fine to support only non-wildcard patterns, but isn't it used elsewhere?

pranavkm: We use it in Razor but that needs to be removed.

pranavkm: Removed the need for locking. We don't cache these tokens any more so we don't have to concern ourselves with concurrency

pranavkm:
Suggestions on who should look at this?

pranavkm:
cc @victorhurdugaci

src/Microsoft.Extensions.FileProviders.Physical/PollingFileChangeToken.cs
```diff
@@ -0,0 +1,65 @@
...
+                {
+                    // Return true, if the change token ever changed in the past. The expecatation is that change tokens
+                    // are not re-used once expired, so the caller must discard this instance once it sees it has changed.
+                    return _hasChanged;
+                }
+
+                var currentTime = DateTime.UtcNow;
+                if (currentTime - _lastCheckedTimeUtc < PollingInterval)
+                {
+                    return false;
```
victorhurdugaci: This code is not threadsafe. There's a chance that two calls will will get true and false if they run around the same time because one updates the _lastCheckedTimeUtc and returns true while the other does this runs this if block and returns false.

You should return _hasChanged instead of false

src/Microsoft.Extensions.FileProviders.Physical/PollingFileChangeToken.cs
```diff
@@ -0,0 +1,65 @@
...
+
+using System;
+using System.IO;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Physical
+{
+    public class PollingFileChangeToken : IChangeToken
+    {
+        private FileInfo _fileInfo;
```
victorhurdugaci: nit. readonly

src/Microsoft.Extensions.FileProviders.Physical/PollingFileChangeToken.cs
```diff
@@ -0,0 +1,65 @@
...
+        private DateTime _previousWriteTimeUtc;
+        private DateTime _lastCheckedTimeUtc;
+        private bool _hasChanged;
+
+        public PollingFileChangeToken(FileInfo fileInfo)
+        {
+            _fileInfo = fileInfo;
+            _previousWriteTimeUtc = GetLastWriteTimeUtc();
```
victorhurdugaci: Is the assumption that the file exists when you create the object? There's a chance that the file could get deleted between the moment you get the FileInfo and the moment this code runs. If that happens, this change token will never fire because the last write time will always be DateTime.MinValue

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
```diff
@@ -137,54 +190,47 @@ private void ReportChangeForMatchedEntries(string pattern)
...
-                filter.Replace('\\', Path.DirectorySeparatorChar) :
-                filter.Replace('/', Path.DirectorySeparatorChar);
-
-            return filter;
-        }
+            Task.Run(() =>
+            {
+                try
+                {
+                    matchInfo.TokenSource.Cancel();
```
victorhurdugaci: Why do you cancel it in a task?

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
```diff
@@ -137,54 +190,47 @@ private void ReportChangeForMatchedEntries(string pattern)
...
 
-        private string NormalizeFilter(string filter)
+        private static string NormalizePath(string filter) => filter.Replace(Path.DirectorySeparatorChar, '/');
+
+        private static bool IsDirectoryPath(string path)
         {
-            // If the searchPath ends with \ or /, we treat searchPath as a directory,
-            // and will include everything under it, recursively.
-            if (IsDirectoryPath(filter))
+            return path.Length > 0
```
victorhurdugaci: I see that the old code was doing a null check too. Did you remove it by mistake?

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
```diff
@@ -33,30 +35,88 @@ public PhysicalFilesWatcher(string root, FileSystemWatcher fileSystemWatcher)
...
+            ChangeTokenInfo tokenInfo;
+            if (!_matchInfoCache.TryGetValue(filePath, out tokenInfo))
+            {
+                var cancellationTokenSource = new CancellationTokenSource();
+                var cancellationChangeToken = new CancellationChangeToken(cancellationTokenSource.Token);
+                tokenInfo = new ChangeTokenInfo(cancellationTokenSource, cancellationChangeToken);
+                tokenInfo = _matchInfoCache.GetOrAdd(filePath, tokenInfo);
+            }
+
+
```
victorhurdugaci: Empty line

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
```diff
@@ -33,30 +35,88 @@ public PhysicalFilesWatcher(string root, FileSystemWatcher fileSystemWatcher)
...
+        private IChangeToken ResolveFileTokensForGlobbingPattern(string filter)
+        {
+            var matcher = new Matcher(StringComparison.OrdinalIgnoreCase);
+            matcher.AddInclude(filter);
+
+            var directoryBase = new DirectoryInfoWrapper(new DirectoryInfo(_root));
+            var result = matcher.Execute(directoryBase);
+
+            var changeTokens = new List<IChangeToken>();
```
victorhurdugaci: Minor perf optimization. You can initialize the list to the result.Files.Count capacity so it doesn't have to resize at all

victorhurdugaci:
:watch:

src/Microsoft.Extensions.FileProviders.Physical/PollingFileChangeToken.cs
pranavkm: There really is no way to verify that is there? If a file is deleted -> created -> deleted between the polling interval, there's no trace to know that it ever existed.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
pranavkm: Existing code from https://github.com/aspnet/FileSystem/pull/183/files#diff-ce3fab8bfc1ed9de2708189277cecb0eL55

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
pranavkm: Added a null check to https://github.com/aspnet/FileSystem/pull/183/files#diff-8bb8e51541651d6e13df23d9a7311bb1R44. The calling code also perfoms a null check so it really isn't a valid input here.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
pranavkm: `result.Files` is an `IEnumerable`, so no `Count` on it.

src/Microsoft.Extensions.FileProviders.Physical/PollingFileChangeToken.cs
victorhurdugaci: no, no. I'm talking about something different. I'm talking about the file being deleted between the moment when you invoke the constructor and the moment when this line of code runs. Basically if the file is deleted and never recreated you'll never know that it was deleted. My suggestion was just to make this as "_hasChanged" if the file doesn't exist when you create the object - it means that it existed just before

pranavkm:
:up: :date:

src/Microsoft.Extensions.FileProviders.Physical/PollingFileChangeToken.cs
pranavkm: The `FileInfo` doesn't necessarily represent a file on disk. It could have `Exists = false`

victorhurdugaci:
:shipit: after you investigate the travis failure

PR #191, opened 2016-05-19
Add test that ensures system files return NotFound

pakrym:
@Tratcher

### DIFF
--- a/test/Microsoft.Extensions.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
+++ b/test/Microsoft.Extensions.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
@@ -110,6 +110,28 @@ namespace Microsoft.Extensions.FileProviders
             }
         }
 
+        [ConditionalFact]
+        [OSSkipCondition(OperatingSystems.Linux, SkipReason = "Hidden and system files only make sense on Windows.")]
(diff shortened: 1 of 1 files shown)


test/Microsoft.Extensions.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -110,6 +110,28 @@ public void GetFileInfoReturnsNotFoundFileInfoForHiddenFile()
...
+        {
+            using (var root = new DisposableFileSystem())
+            {
+                using (var provider = new PhysicalFileProvider(root.RootPath))
+                {
+                    var fileName = Guid.NewGuid().ToString();
+                    var filePath = Path.Combine(root.RootPath, fileName);
+                    File.Create(filePath);
+                    var fileInfo = new FileInfo(filePath);
+                    File.SetAttributes(filePath, fileInfo.Attributes | FileAttributes.System);
```
Tratcher: These should be tested separately, otherwise you couldn't tell if one was broken.

pakrym: Which ones? This line only adds `FileAttributes.System` to existing attributes, there is another test for hidden.

Tratcher: Ah, nevermind.

Tratcher:
:shipit:

Issue #145, opened 2015-11-24
Suggest to make NotFoundFileInfo and NoopChangeToken public


PrimeObjects:

Eilon:
Can you clarify the reason for wanting them public? Are you writing a custom file system provider? Or something else?

PrimeObjects:
I'm writing a ViewFileProvider, I hope this FileProvider can serve MVC by using view files saved in the database.

I need to return NoopChangeToken and NotFoundFileInfo like what RazorEmbeddedFileProvider did.
I want to use the one defined in .NET in my class directly.
e.g.
public IChangeToken Watch(string pattern)
        {
            return NoopChangeToken.Singleton;
        }

Eilon:
Ah, got it. So just for code re-use, which sounds fine. Just wanted to make sure it wasn't for checking equality, which would be wrong (e.g. checking `if (foo.GetType() == typeof(NoopChangeToken))` and making some decision based on that.

PrimeObjects:
thank you for reminding. I cloned the two classes into my namespace for now. if you decide to make the two classes public I should be able use them directly. 

Thanks again.
Gary

mgrosperrin:
Hi,
maybe we could provide a factory to create _empty_ implementation of these types (`IFileInfo`, `IDirectoryContents` and `IChangeToken`). This would protect against comparing the type (except if people do scary things like `foo.GetType() == factory.GetXXX().GetType()`) and allow reuse of the types.
I can provide you a PR to prototype it if needed (I just need a name for the project hosting the factory :smile:).

pranavkm:
@PrimeObjects all of these packages are in the `Microsoft.AspNet.FileProviders.Sources` package. You should be able to reference them and use that.

mgrosperrin:
@pranavkm I have tried to do this for my composite file provider, but the `Microsoft.AspNet.FileProviders.Sources` package has not been published in the RC1 wave. Will it be published as part of RC2 (it is in the nightly) ?

pranavkm:
We don't publish any of the Sources packages to NuGet.org. You'd have to pick it up from `aspnetmaster`.

Driedas:
[...]

PR #195, opened 2016-05-25
expose NoopChangeToken, NotFoundDirectoryContents, NotFoundFileInfo as public

Driedas:
changes as per the discussion on issue https://github.com/aspnet/FileSystem/issues/145

unit tests passing (executed via dotnet test)

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Abstractions/NotFoundDirectoryContents.cs
+++ b/src/Microsoft.Extensions.FileProviders.Abstractions/NotFoundDirectoryContents.cs
@@ -7,7 +7,10 @@ using System.Linq;
 
 namespace Microsoft.Extensions.FileProviders
 {
-    internal class NotFoundDirectoryContents : IDirectoryContents
+    /// <summary>
+    /// Represents a non-existing directory
+    /// </summary>
+    public class NotFoundDirectoryContents : IDirectoryContents
     {
         public bool Exists => false;
 
--- a/src/Microsoft.Extensions.FileProviders.Abstractions/NotFoundFileInfo.cs
+++ b/src/Microsoft.Extensions.FileProviders.Abstractions/NotFoundFileInfo.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Extensions.FileProviders
     /// <summary>
     /// Represents a non-existing file.
     /// </summary>
-    internal class NotFoundFileInfo : IFileInfo
+    public class NotFoundFileInfo : IFileInfo
     {
         public NotFoundFileInfo(string name)
         {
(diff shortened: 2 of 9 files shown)


src/Microsoft.Extensions.FileProviders.Abstractions/NoopChangeToken.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Extensions.FileProviders
 {
-    internal class NoopChangeToken : IChangeToken
+    public class NoopChangeToken : IChangeToken
```
pranavkm: Naming concern since we're making this type public - should this be called `NullChangeToken`? We name other things similarly e.g. `NullLogger` for things that don't do much. Thoughts @Eilon ?

src/Microsoft.Extensions.FileProviders.Abstractions/NotFoundDirectoryContents.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.Extensions.FileProviders
 {
-    internal class NotFoundDirectoryContents : IDirectoryContents
+    public class NotFoundDirectoryContents : IDirectoryContents
```
pranavkm: Doc comment:

>  A non-existing directory

src/Microsoft.Extensions.FileProviders.Abstractions/NoopChangeToken.cs
Eilon: Yeah I suppose this is basically the Null Pattern.

Eilon:
@pranavkm do we need a bunch of reaction PRs in other repos for this? I suspect this might be a slightly destabilizing change due to other repos needing to react to this.

Eilon:
Oh actually I wonder if perhaps this is totally safe because the type is moving from being internal in `.Sources` to being public in the main package. So presumably it's just a switcheroo and perhaps safe...

pranavkm:
Yeah, the move should be fine and the rename too since it doesn't look like it's used outside of this repo.

Driedas:
@pranavkm renamed the NoopChangeToken to NullChangeToken and added the comment for NotFoundDirectoryContents, as per your comments
Have tried to come up with a reasonable comment for the NullChangeToken as well, that didn't involve the word "fake" or "stub", as these are mostly associated with *testing, but I have nothing better than "An empty change token that doesn't raise any change callbacks"

pranavkm:
@Driedas that sounds good.

pranavkm:
Let me know when you've have added the resx string and I'll merge this in.

Driedas:
@pranavkm comment added
Do you want me to squash the commits into a single one or is ok like this?

pranavkm:
Squashing it would be great.

Driedas:
done...

pranavkm:
Thanks!

Eilon:
@Driedas thanks for the PR!

Issue #190, opened 2016-05-11
Default EmbeddedFileProvider baseNamespace parameter to assembly name


pakrym:
Lets take assembly named `MyCompany.MyApplication` with embedded folder `wwwroot` that has `text.txt` file in it.

When using `EmbeddedFileProvider.ctor(Assembly assembly)` it will set `baseNamespace` to empty by default, so URL will look like: `http://localhost/MyCompany/MyApplication/wwwroot/text.txt`.

If we change it's behavior to set `baseNamespace` to assembly name it will look like `http://localhost/wwwroot/text.txt`, which in my opinion is more desirable.

In `dotnet cli` namespace for embedded files defaults to project name, and assembly name default's to it too, so this change would cover most of cli use cases.

muratg:
Do we currently support multiple assemblies for the file provider?  If we go with this proposal, and and if there's a clash in the folder name, what happens?

muratg:
Ping

muratg:
@pakrym Let's make this change in RTM. Once done make sure to post an announcement.

PR #196, opened 2016-05-26
Set embedded files provider namespace to assembly name by default

pakrym:
Fixes: https://github.com/aspnet/FileSystem/issues/190
@pranavkm @BrennanConroy


BrennanConroy:
:shipit:

Issue #201, opened 2016-06-09
Long path or filename throws unhandled exception


Tratcher:
_From @cweb on June 9, 2016 23:16_

A long request path will throw an unhandled exception.

```
An unhandled exception has occurred: The specified path, file name, or both are too long. The fully quali
fied file name must be less than 260 characters, and the directory name must be less than 248 characters.
System.IO.PathTooLongException: The specified path, file name, or both are too long. The fully qualified file name must
be less than 260 characters, and the directory name must be less than 248 characters.
   at System.IO.PathHelper.GetFullPathName()
   at System.IO.Path.NormalizePath(String path, Boolean fullCheck, Int32 maxPathLength, Boolean expandShortPaths)
   at System.IO.Path.GetFullPathInternal(String path)
   at Microsoft.Extensions.FileProviders.PhysicalFileProvider.GetFullPath(String path)
   at Microsoft.Extensions.FileProviders.PhysicalFileProvider.GetFileInfo(String subpath)
   at Microsoft.AspNetCore.StaticFiles.StaticFileContext.LookupFileInfo() in C:\Users\casaba\Documents\StaticFiles\src\M
icrosoft.AspNetCore.StaticFiles\StaticFileContext.cs:line 141
   at Microsoft.AspNetCore.StaticFiles.StaticFileMiddleware.Invoke(HttpContext context) in C:\Users\casaba\Documents\Sta
ticFiles\src\Microsoft.AspNetCore.StaticFiles\StaticFileMiddleware.cs:line 86
   at Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware.<Invoke>d__6.MoveNext()
```

_Copied from original issue: aspnet/StaticFiles#138_

PR #203, opened 2016-06-10
Treat Path.GetFullPath exceptions as unknown files

JunTaoLuo:
#201 #204 

cc @Tratcher which milestone should this go into? I might also add some tests for this.


Tratcher:
@muratg @Eilon recommend this for 1.0.
:shipit:

muratg:
I recommend sending an email for approval if we want to take this in 1.0.0

Issue #205, opened 2016-06-20
PhysicalFilesWatcher should defend against PathTooLongException


Eilon:
In the various event handlers in [PhysicalFilesWatcher](https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs), there can be exceptions thrown for paths that are too long. For example, constructing a `FileInfo` with a long path can throw.

Because this code runs in background threads, this can easily take down a process that's using it.

@NTaylorMullen did some related fixes in https://github.com/aspnet/FileSystem/issues/187, but it seems that we need to handle more cases.

See Watson info here: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workitems?id=186004&_a=edit

BrennanConroy:
https://github.com/aspnet/FileSystem/commit/b58bbe011ab88335d532762740c43381b36244a7

Issue #189, opened 2016-05-10
Revisit how physical file watcher handles events


NTaylorMullen:
We did a quick fix for https://github.com/aspnet/FileSystem/issues/187 for RC2 but we should revisit the solution for RTM. Aka, it's bad that we're swallowing exceptions and should investigate other alternatives.

natemcmaster:
@muratg I recommend we close as wont-fix. The result of this bug is that a file watch token may not always fire for renamed events when the rename caused the directory structure to change permissions or exceed the MAX_PATH length.

I've looked at the changes made in #187, and there aren't good alternatives. In this specific case, it is the System.IO API that is throws when there is an issue traversing the directory structure. There aren't any non-throwing System.IO apis we can use to achieve the same result.

muratg:
👍

PR #207, opened 2016-06-28
Catch in background threads

BrennanConroy:
#205 
@Eilon @NTaylorMullen 

I know about #189 but this will be here if we don't want to do anything else.


NTaylorMullen:
:shipit:

Issue #208, opened 2016-06-29
EmbeddedFileProvider - Change _lastModified date to last write time of an assembly


martinba1:
Could you please consider change of _lastModified date to last write time of an assembly? This information would be very useful if you want to use this file provider also for different purposes than Razor View.

FileSystem/src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
Current: _lastModified = DateTimeOffset.MaxValue;

My idea is:
if (assembly.Location != null)
    _lastModified = new FileInfo(assembly.Location).LastWriteTime;
else
    _lastModified = DateTimeOffset.MaxValue; 

Thank you

muratg:
Agreed.

BrennanConroy:
https://github.com/aspnet/FileSystem/commit/a5366af6614835ea4faff7e35c58604563ba5908

PR #210, opened 2016-06-30
Set modified date in EmbeddedFileProvider

BrennanConroy:
#208 
Apps that target `>=net462` or `netcoreapp1.0` will get the new behavior
Other targets will get the `DateTimeOffset.Now` behavior

@Eilon @Tratcher @pakrym

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
+++ b/src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
@@ -49,8 +49,24 @@ namespace Microsoft.Extensions.FileProviders
 
             _baseNamespace = string.IsNullOrEmpty(baseNamespace) ? string.Empty : baseNamespace + ".";
             _assembly = assembly;
-            // REVIEW: Does this even make sense?
-            _lastModified = DateTimeOffset.MaxValue;
+
+            _lastModified = DateTimeOffset.UtcNow;
+
+#if NETSTANDARD1_5 || NET451
+            if (!string.IsNullOrEmpty(_assembly.Location))
+            {
+                try
+                {
+                    _lastModified = File.GetLastWriteTimeUtc(_assembly.Location);
+                }
+                catch (PathTooLongException)
+                {
+                }
+                catch (UnauthorizedAccessException)
+                {
+                }
+            }
+#endif
         }
 
         /// <summary>
--- a/src/Microsoft.Extensions.FileProviders.Embedded/project.json
+++ b/src/Microsoft.Extensions.FileProviders.Embedded/project.json
@@ -24,13 +24,18 @@
       "version": "1.1.0-*",
       "type": "build"
     },
-    "Microsoft.Extensions.FileProviders.Abstractions": "1.1.0-*"
+    "Microsoft.Extensions.FileProviders.Abstractions": "1.1.0-*",
+    "System.Runtime.Extensions": "4.1.0-*"
   },
   "frameworks": {
-    "netstandard1.0": {
+    // need to keep netstandard1.0 until ASP.NET Core 2.0 because it is a breaking change if we remove it
+    "netstandard1.0": { },
+    "netstandard1.5": {
       "dependencies": {
-        "System.Runtime.Extensions": "4.1.0-*"
+        "System.IO.FileSystem": "4.0.1-*",
+        "System.Reflection": "4.1.0-*"
       }
-    }
+    },
+    "net451": { }
   }
 }


src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -49,8 +49,15 @@ public EmbeddedFileProvider(Assembly assembly, string baseNamespace)
 
             _baseNamespace = string.IsNullOrEmpty(baseNamespace) ? string.Empty : baseNamespace + ".";
             _assembly = assembly;
-            // REVIEW: Does this even make sense?
-            _lastModified = DateTimeOffset.MaxValue;
+
+            _lastModified = DateTimeOffset.Now;
```
pranavkm: `UtcNow`?

src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -49,8 +49,15 @@ public EmbeddedFileProvider(Assembly assembly, string baseNamespace)
...
+
+#if NETSTANDARD1_5
+            if (_assembly.Location != null)
+            {
+                _lastModified = new FileInfo(_assembly.Location).LastWriteTime;
```
pranavkm: `File.GetLastWriteTimeUtc`

BrennanConroy:
🆙 📅

src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
```diff
@@ -49,8 +49,15 @@ public EmbeddedFileProvider(Assembly assembly, string baseNamespace)
...
             _assembly = assembly;
-            // REVIEW: Does this even make sense?
-            _lastModified = DateTimeOffset.MaxValue;
+
+            _lastModified = DateTimeOffset.UtcNow;
+
+#if NETSTANDARD1_5
+            if (_assembly.Location != null)
+            {
+                _lastModified = File.GetLastWriteTimeUtc(_assembly.Location);
```
Tratcher: This can throw a lot of exceptions that we may be better off suppressing.
https://msdn.microsoft.com/en-us/library/system.io.file.getlastwritetimeutc(v=vs.110).aspx

src/Microsoft.Extensions.FileProviders.Embedded/project.json
```diff
@@ -24,12 +24,15 @@
...
       "type": "build"
     },
-    "Microsoft.Extensions.FileProviders.Abstractions": "1.1.0-*"
+    "Microsoft.Extensions.FileProviders.Abstractions": "1.1.0-*",
+    "System.Runtime.Extensions": "4.1.0-*"
   },
   "frameworks": {
-    "netstandard1.0": {
+    "netstandard1.0": { },
+    "netstandard1.5": {
```
Tratcher: We can add comments again, right? This might be worth a comment.

Tratcher:
:shipit:

src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
BrennanConroy: Hmm, thoughts @Eilon @pranavkm

pakrym: :+1: 

> The location of the loaded file that contains the manifest. If the loaded file was shadow-copied, the location is that of the file after being shadow-copied. If the assembly is loaded from a byte array, such as when using the Load(Byte[]) method overload, the value returned is an empty string ("").

Lots of edge cases.

pranavkm: No harm in try catching this? On a bit of a tangent, did we consider having the alternative where the user specifies this value? It removes a bunch of this guessing from our code (plus the oddity with having different behavior based on what you might be targeting)?

Tratcher: That's not a bad cop-out..

BrennanConroy: We considered it, but weren't huge fans

pranavkm: casual fans?

pranavkm:
:shipit:

src/Microsoft.Extensions.FileProviders.Embedded/project.json
pakrym: Also we possibly want to cross compile to net451 to get this functionality on full framework.

BrennanConroy: That is a good point, will add

BrennanConroy:
🆙 📅

BrennanConroy:
Ping @Tratcher @pranavkm

pranavkm:
:shipit:

Issue #206, opened 2016-06-27
IFileProvider use case and abstraction issues


dazinator:
Questioning whether the `IFileProvider` abstraction needs improving (or its implementations) - due to following issues. I am finding usage of the abstraction needs to be different for different providers, so in essence you have to cater for all specific implementations and the IFileProvider abstraction isn't helpful. 

Example use case.. 

I want an RC2 application to list all available CSS files that are themes for the application, so the user can select one and apply a theme to the site.

In order to distinguish a theme file from other files, the rule I am using is, it must live in a Css/Themes directory.
However, in addition to supporting just physical files (from the web applications wwwroot/css/themes directory), the application should also support the ability for theme files to live within referenced projects within the solution.

Here is an example solution directory structure:

WebApplication/wwwroot/css/themes/theme1.css
WebApplication/wwwroot/css/themes/theme2.css
ModuleA/wwwroot/css/themes/moduleA.css
ModuleB/wwwroot/css/themes/moduleB.css

In order to support files located in referenced projects, [it means actually embedding the files into the assembly](https://github.com/aspnet/Home/issues/1564) and using an EmbeddedFileProvider over those assemblies. CopyToOutput is useless [because its non transitive](https://github.com/dotnet/cli/issues/2902) 

So to fulfil the use case, the application starts and creates an "EmbeddedFileProvider" for the referenced project assemblies, pointing to their "wwwroot" folders, as well as a PhysicalFileProvider for the main web applications "wwwroot" directory, and then wraps those in a CompositeFileProvider to give one IFileProvider instance that we can work with to query for Theme files..

Now given the `IFileProvider` instance, which is `CompositeFileProvider` we want to list the themes css files. If things were aligned it would be as simple as:

```
var allThemesFiles = fileProvider.GetDirectoryContents("css/themes")
```

[...]

PR #211, opened 2016-07-15
Support of non-flat tree in the Embedded Resource File Provider.

Yves57:
See #206.
- Added a parameter `isDirectory` in the split function to be able to have specific processing depending on the path is a file or a directory.
- Keeped the current behavior when there is no 'splitter'. But of course in this case it keeps the difference between `GetFileInfo()` that supports more or less subpaths while `GetDirectoryContents()` does not support them.

Pending:
- The `CreateFromFileExtensions()` helper method. I wait your validation on the first PR.

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
+++ b/src/Microsoft.Extensions.FileProviders.Embedded/EmbeddedFileProvider.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Extensions.FileProviders
         private readonly Assembly _assembly;
         private readonly string _baseNamespace;
         private readonly DateTimeOffset _lastModified;
+        private readonly Dictionary<string, IFileInfo> _allEntries;         // Values may be 'null' if information must be dynamically calculated
+        private readonly Dictionary<string, List<IFileInfo>> _directories;  // Values are directory children
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EmbeddedFileProvider" /> class using the specified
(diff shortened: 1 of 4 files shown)


Yves57:
What's the error in Travis? Do I something wrong?

dazinator:
Just a note to mention that you can now also achieve this using my InMemoryFileProvider located here: https://github.com/dazinator/Dazinator.AspNet.Extensions.FileProviders

Here is a test that shows creating an `InMemoryFileProvider`, and adding an `EmbeddedFileInfo` to be resolved on the subpath: `/somefolder/folder/myfile.txt`.


```

 public class EmbeddedFileInfoTests
    {

        [Fact]
        public void GetFileInfo_ReturnsFileInfo_ForEmbeddedFile()
        {
            using (var provider = new InMemoryFileProvider())
            {
                var embeddedFileInfo = new EmbeddedFileInfo(GetAssemblyFromType(this.GetType()),
                    $"Dazinator.AspNet.Extensions.FileProviders.Tests.Resources.myresource.txt",
                    "myfile.txt");

                provider.Directory.AddFile("/some/folder", embeddedFileInfo);

                var info = provider.GetFileInfo("/some/folder/myfile.txt");
                using (var stream = info.CreateReadStream())
                {
                    Assert.NotNull(stream);
                    Assert.True(stream.Length > 0);
                }

            }
        }


        public static Assembly GetAssemblyFromType(Type type)
        {

#if NETSTANDARD
            var assy = type.GetTypeInfo().Assembly;
#else
            var assy = type.Assembly;
#endif
            return assy;
        }



    }

```

natemcmaster:
@muratg this was added to Backlog but left this open. Do we want to take this change or should we close this now?

muratg:
Closing this. Thanks @Yves57 and @dazinator !

PR #214, opened 2016-07-19
Reduce allocation in PhysicalFileProvider when used in StaticFiles

pakrym:
By default we pass 64k buffer size but it causes significant overhead for small sized files.
## Before:
Using 200byte file

## After:

@Tratcher  
@davidfowl ` // Note: Buffer size must be greater than zero, even if the file size is zero.` why?

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/FileSystemInfoHelper.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/FileSystemInfoHelper.cs
@@ -14,8 +14,8 @@ namespace Microsoft.Extensions.FileProviders
                 return true;
             }
             else if (fileSystemInfo.Exists &&
-                (fileSystemInfo.Attributes.HasFlag(FileAttributes.Hidden) ||
-                fileSystemInfo.Attributes.HasFlag(FileAttributes.System)))
+                ((fileSystemInfo.Attributes & FileAttributes.Hidden) != 0 ||
+                 (fileSystemInfo.Attributes & FileAttributes.System) != 0))
             {
                 return true;
             }
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
@@ -8,6 +8,11 @@ namespace Microsoft.Extensions.FileProviders.Physical
 {
     public class PhysicalFileInfo : IFileInfo
     {
+        private const int MaxBufferSize = 1024 * 64;
+        // Note: Buffer size must be greater than zero, even if the file size is zero
+        // otherwise FileStream constructor will throw
+        private const int MinBufferSize = 1;
+
         private readonly FileInfo _info;
 
         public PhysicalFileInfo(FileInfo info)
@@ -29,8 +34,7 @@ namespace Microsoft.Extensions.FileProviders.Physical
 
         public Stream CreateReadStream()
         {
-            // Note: Buffer size must be greater than zero, even if the file size is zero.
-            var bufferSize = (int)Math.Min(1024 * 64, Math.Max(1, Length));
+            var bufferSize = (int)Math.Min(MaxBufferSize, Math.Max(MinBufferSize, Length));
             return new FileStream(
                 PhysicalPath,
                 FileMode.Open,
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
@@ -30,12 +30,13 @@ namespace Microsoft.Extensions.FileProviders.Physical
         public Stream CreateReadStream()
         {
             // Note: Buffer size must be greater than zero, even if the file size is zero.
+            var bufferSize = (int)Math.Min(1024 * 64, Math.Max(1, Length));
             return new FileStream(
                 PhysicalPath,
                 FileMode.Open,
                 FileAccess.Read,
                 FileShare.ReadWrite,
-                Math.Min((int)Length, 1024 * 64),
+                bufferSize,
                 FileOptions.Asynchronous | FileOptions.SequentialScan);
         }
     }
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Extensions.FileProviders.Physical
                 FileMode.Open,
                 FileAccess.Read,
                 FileShare.ReadWrite,
-                1024 * 64,
+                Math.Min((int)Length, 1024 * 64),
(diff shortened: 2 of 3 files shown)


src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
```diff
@@ -35,7 +35,7 @@ public Stream CreateReadStream()
                 FileMode.Open,
                 FileAccess.Read,
                 FileShare.ReadWrite,
-                1024 * 64,
+                Math.Min((int)Length, 1024 * 64),
```
pranavkm: `FileStream` defines the default buffer size to be 4k. Any idea why we picked a significantly larger number? Does it make sense to stick to the 4k value?

pakrym: Maybe https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/Win32FileStream.cs#L1112 ?

Tratcher: The fewer round trips we can do to the disk, the better.

Tratcher:
:shipit:

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
```diff
@@ -29,13 +29,12 @@ public PhysicalFileInfo(FileInfo info)
 
         public Stream CreateReadStream()
         {
-            // Note: Buffer size must be greater than zero, even if the file size is zero.
```
campersau: Was this comment invalid? Because now `Math.Min((int)Length, 1024 * 64)` can return `0` which would result in an exception. Or can't `Length` be `0`? https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/FileStream.cs#L85

pakrym: I thought I verified 0 length file and it worked, but I re-tested and it didn't. Thank you for noticing.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
```diff
@@ -29,13 +29,12 @@ public PhysicalFileInfo(FileInfo info)
...
             return new FileStream(
```
rynowak: Should we use the buffer pool here instead? Then we can use whatever sized buffer we like.

src/Microsoft.Extensions.FileProviders.Physical/FileSystemInfoHelper.cs
```diff
@@ -14,8 +14,8 @@ internal static bool IsHiddenFile(FileSystemInfo fileSystemInfo)
                 return true;
             }
             else if (fileSystemInfo.Exists &&
-                (fileSystemInfo.Attributes.HasFlag(FileAttributes.Hidden) ||
-                fileSystemInfo.Attributes.HasFlag(FileAttributes.System)))
```
rynowak: Yes, purge all use of `Enum.HasFlags`, it is evil.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
campersau: What if `Length` is greater than `int.MaxValue`? I think it will overflow and would result in `-1` which would also throw. https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/FileStream.cs#L85

```c#
long lng = long.MaxValue;
var i = Math.Min((int)lng, 1);
// i == -1
```

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
pakrym: It's not us, it's FileStream implementation.

test/Microsoft.Extensions.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -114,6 +114,23 @@ public void GetFileInfoReturnsNotFoundFileInfoForRelativePathWithEmptySegmentsTh
...
+        [Fact]
+        public void CreateReadStreamSuccedsOnEmptyFile()
+        {
+            using (var root = new DisposableFileSystem())
+            {
+                using (var provider = new PhysicalFileProvider(root.RootPath))
+                {
+                    var fileName = Guid.NewGuid().ToString();
+                    var filePath = Path.Combine(root.RootPath, fileName);
+                    using (File.Create(filePath)) { }
```
pranavkm: `File.WriteAllBytes(filePath, new byte[0])`? A bit cleaner than the single line dispose.

test/Microsoft.Extensions.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -114,6 +114,23 @@ public void GetFileInfoReturnsNotFoundFileInfoForRelativePathWithEmptySegmentsTh
...
+                    var info = provider.GetFileInfo(fileName);
+                    var stream = info.CreateReadStream();
```
pranavkm: `using`

pranavkm:
:shipit:

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
rynowak: ah right, I remember now why I didn't just suggest that in the bug.

rynowak:
:shipit:

davidfowl:
Looks good

davidfowl:
Can you do some tests with large files to see what the before/after looks like?

pakrym:
@davidfowl why would it change for large files?

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
```diff
@@ -30,12 +30,13 @@ public PhysicalFileInfo(FileInfo info)
         public Stream CreateReadStream()
         {
             // Note: Buffer size must be greater than zero, even if the file size is zero.
+            var bufferSize = (int)Math.Min(1024 * 64, Math.Max(1, Length));
```
davidfowl: 1024 * 64 a constant MaxBufferSize

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
```diff
@@ -30,12 +30,13 @@ public PhysicalFileInfo(FileInfo info)
         public Stream CreateReadStream()
         {
             // Note: Buffer size must be greater than zero, even if the file size is zero.
```
MartinJohns: Please add a comment **why** this is the case.

Issue #215, opened 2016-07-21
Idea to minimize buffer allocations in PhysicalFileProvider


davidfowl:
Credit goes to @stephentoub 

Pass the minimum buffer size (1) to the FileStream constructor and have callers use pooled buffers for reading.

/cc @pakrym @Tratcher @rynowak

pakrym:
Would read skipping internal buffer in this case?

davidfowl:
That's the idea.

pakrym:
I'll try and profile it tomorrow.

pakrym:
Would be nice to allow passing 0 to `bufferSize` then.

stephentoub:
While it may be nice aesthetically, it shouldn't really impact anything.  The buffer shouldn't be allocated (at least not in corefx), as the buffer is lazily-initialized and will be skipped if the number of bytes to be read is >= bufferSize.

muratg:
@pakrym did it help (if you got the chance to profile)?

pakrym:
Yes, I implemented it. Buffer allocations from `FileStream` are not a problem anymore.

PR #216, opened 2016-07-21
Prevent buffer allocation in FileStream returned from PhysicalFileInfo.CreateReadStream

pakrym:
Part of https://github.com/aspnet/FileSystem/issues/215

@pranavkm @davidfowl

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
@@ -8,11 +8,6 @@ namespace Microsoft.Extensions.FileProviders.Physical
 {
     public class PhysicalFileInfo : IFileInfo
     {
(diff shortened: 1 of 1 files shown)


davidfowl:
:shipit:

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs
```diff
@@ -34,7 +29,9 @@ public PhysicalFileInfo(FileInfo info)
 
         public Stream CreateReadStream()
         {
-            var bufferSize = (int)Math.Min(MaxBufferSize, Math.Max(MinBufferSize, Length));
+            // We are setting buffer size to 1 to prevent FileStream from allocating it's internal buffer
+            // 0 causes constructor to throw
+            var bufferSize = 1;
```
davidfowl: This could be an argument to to physical file system provider. That way if somebody wants more control they can get that via this argument.

PR #220, opened 2016-08-02
[WIP] FileInfo caching provider

pakrym:
@davidfowl

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
@@ -1,20 +1,28 @@
-﻿using Microsoft.CodeAnalysis.InternalUtilities;
+﻿using System;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Extensions.FileProviders.Physical
 {
-    public class CachingFileProvider: IFileProvider
+    public class CachingFileProvider: IFileProvider, IDisposable
     {
         private readonly IFileProvider _baseProvider;
         private readonly ConcurrentLruCache<string, IFileInfo> _fileCache;
         private readonly ConcurrentLruCache<string, IDirectoryContents> _directoryCache;
+        private Func<string, IFileInfo> _getFileInfo;
+        private Func<string, IDirectoryContents> _getDirectoryContents;
+        private IDisposable _changeCallbackRegistration;
 
-        public CachingFileProvider(IFileProvider baseProvider, int fileCapacity, int directoryCapacity)
+        public CachingFileProvider(IFileProvider baseProvider, int fileCapacity, int directoryCapacity, string watchFilter)
         {
             _baseProvider = baseProvider;
             _fileCache = new ConcurrentLruCache<string, IFileInfo>(fileCapacity);
             _directoryCache = new ConcurrentLruCache<string, IDirectoryContents>(directoryCapacity);
-            _baseProvider.Watch("/*").RegisterChangeCallback(Invalidate, null);
+            if (watchFilter != null)
+            {
+                _changeCallbackRegistration = _baseProvider.Watch(watchFilter).RegisterChangeCallback(Invalidate, null);
+            }
+            _getDirectoryContents = _baseProvider.GetDirectoryContents;
+            _getFileInfo = _baseProvider.GetFileInfo;
         }
 
         private void Invalidate(object obj)
@@ -25,17 +33,22 @@ namespace Microsoft.Extensions.FileProviders.Physical
 
         public IFileInfo GetFileInfo(string subpath)
         {
-            return _fileCache.GetOrAdd(subpath, subpath, _baseProvider.GetFileInfo);
+            return _fileCache.GetOrAdd(subpath, subpath, _getFileInfo);
         }
 
         public IDirectoryContents GetDirectoryContents(string subpath)
         {
-            return _directoryCache.GetOrAdd(subpath, subpath, _baseProvider.GetDirectoryContents);
+            return _directoryCache.GetOrAdd(subpath, subpath, _getDirectoryContents);
         }
 
         public IChangeToken Watch(string filter)
         {
             return _baseProvider.Watch(filter);
+        }
+
+        public virtual void Dispose()
+        {
+            _changeCallbackRegistration?.Dispose();
         }
     }
 }
--- a/src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
@@ -1,0 +1,41 @@
+﻿using Microsoft.CodeAnalysis.InternalUtilities;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Physical
+{
+    public class CachingFileProvider: IFileProvider
+    {
+        private readonly IFileProvider _baseProvider;
+        private readonly ConcurrentLruCache<string, IFileInfo> _fileCache;
+        private readonly ConcurrentLruCache<string, IDirectoryContents> _directoryCache;
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+        private IDisposable _changeCallbackRegistration;
+
+        public CachingFileProvider(IFileProvider baseProvider, int fileCapacity, int directoryCapacity, string watchFilter)
+        {
+            _baseProvider = baseProvider;
+            _fileCache = new ConcurrentLruCache<string, IFileInfo>(fileCapacity);
+            _directoryCache = new ConcurrentLruCache<string, IDirectoryContents>(directoryCapacity);
+            if (watchFilter != null)
+            {
+                _changeCallbackRegistration = _baseProvider.Watch(watchFilter).RegisterChangeCallback(Invalidate, null);
```
pranavkm: This could potentially be polling in the background (which does GetFileInfo) so relying on this does not necessarily help. Additionally adding any of this to a static in time provider (either an embedded file provider or a physical file system that does not change) is probably going to be more overhead. Should we describe what scenarios we're trying to address here? Caching for the file system is too broad a concept and maybe we need something more fine grained?

src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Physical
+{
+    public class CachingFileProvider: IFileProvider, IDisposable
+    {
+        private readonly IFileProvider _baseProvider;
+        private readonly ConcurrentLruCache<string, IFileInfo> _fileCache;
```
pranavkm: Why not use a MemoryCache here?

src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
pakrym: We are trying to decrease amount of `GetFileAttributesEx` calls `PhysicalFileProvider` does and improve `StaticFiles` performance. User that adds this provider should know why is he doing it, Also passing `null` to `watchFilter` disables watcher.

src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
pakrym: I wanted fixed size cache for this one, but can switch to using `MemoryCache`

src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Physical
+{
+    public class CachingFileProvider: IFileProvider, IDisposable
```
MartinJohns: Please add documentation.

davidfowl: Its a WIP (Work in progress) 😄

src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public IChangeToken Watch(string filter)
+        {
+            return _baseProvider.Watch(filter);
+        }
+
+        public virtual void Dispose()
+        {
+            _changeCallbackRegistration?.Dispose();
+        }
```
Yves57: The `_baseProvider` is not disposed? Is it the user that is in charge of disposing it?
Maybe clean the both caches?

davidfowl: IFileProvider isn't disposable.

src/Microsoft.Extensions.FileProviders.Physical/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,247 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Expects non-empty cache. Does not lock.
+        /// </summary>
+        private void UnsafeEvictLastNode()
+        {
+            var lastNode = _nodeList.Last;
+            _nodeList.Remove(lastNode);
```
Yves57: '_nodeList.RemoveLast()'? Otherwise it will search in the whole list to find the last.

src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
Yves57: `PhysicalFileProvider` is disposable. But I suppose that you consider that it is the user job because the interface itself is not disposable.

davidfowl: I'm not even sure where we dispose it in normal cases. It's not in the DI container so it's possible we have a leak in some cases

src/Microsoft.Extensions.FileProviders.Physical/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,247 @@
...
+            lock (_lockObject)
+            {
+                V result;
+                if (UnsafeTryGetValue(key, out result))
+                {
+                    return result;
+                }
+                else
+                {
+                    var value = creator();
```
Yves57: `Creator()` is called inside the Lock. So the File Provider is virtually freezed as long as it has not the information. You don't think it will create a bottleneck in case of simultaneous requests?

src/Microsoft.Extensions.FileProviders.Physical/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,247 @@
...
+        private readonly int _capacity;
+
+        private struct CacheValue
+        {
+            public V Value;
+            public LinkedListNode<K> Node;
+        }
+
+        private readonly Dictionary<K, CacheValue> _cache;
+        private readonly LinkedList<K> _nodeList;
```
Yves57: Just some remarks about `_nodeList`, maybe stupid. Sorry if it is the case...
I would remove `_nodeList`:
- The `LinkedListNode` will create memory fragmentation.
- Removing the older nodes is not a high priority task, it can be delegated to a background task, like in the Memory Cache.
- It will save time because the `GetOrAdd()` won't have to move the node to top each time.
- I'm not sure, but if you do this you may remove `_lockObject` too, and keep only a high performance `ConcurrentDictionary` in `_cache`. It would create some collisions in rare cases (two requests want to add the same value to the cache), but I don't think that its a problem.

src/Microsoft.Extensions.FileProviders.Physical/CachingFileSystemProvider.cs
Yves57: I have no argument against that :-).

src/Microsoft.Extensions.FileProviders.Physical/ConcurrentLruCache.cs
Yves57: I hope not to waste your time, but I send you a source code of a `ConcurrentLruCache` with only a concurrent dictionary. I think it minimizes the bottlenecks.

pakrym: I took `ConcurrentLruCache` from roslyn repo just as a proof of concept, we are yet to decide best data structure to use here and profile to see if this implementation needs improvements.

Issue #194, opened 2016-05-24
File globbing should support pattern checks


victorhurdugaci:
Today you cannot check if if a path matches a pattern without executing the globbing pattern (aka going to disk). This slows down dotnet watch a lot because on every file change it has to go to the disk, fetch all files and the see if the changed file is in the resulting list.

pranavkm:
Isn't this what FileSystemGlobber does - https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs#L107-L108

victorhurdugaci:
Nope. What I want to do is, given:

1. Base path: `/foo`
2. Pattern: `*.js`

Tell me if `/foo/bar.js` matches that without going to the disk.

dazinator:
> Tell me if /foo/baz/bar.js matches that without going to the disk.

Yes please! Does globbing only work with `PhysicalFileProvider` currently then?

I'd like to take:
1. A pattern `**/*.js`
2. An `IEnumerable<string>` paths - `{"some/path/file1.js", "some/path/file2.js", "someother/file/one.txt"}`  

Tell me what the matches are (without going to disk)

BrennanConroy:
https://github.com/aspnet/FileSystem/commit/767f07ff39ec66e567d73a034f11bda260ab6123

PR #222, opened 2016-08-03
Add ability to match pattern without going to disk

BrennanConroy:
https://github.com/aspnet/FileSystem/issues/194

cc: @pakrym

### DIFF
--- a/src/Microsoft.Extensions.FileSystemGlobbing/InMemoryDirectoryInfo.cs
+++ b/src/Microsoft.Extensions.FileSystemGlobbing/InMemoryDirectoryInfo.cs
@@ -1,0 +1,164 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
+using Microsoft.Extensions.FileSystemGlobbing.Internal;
+
+namespace Microsoft.Extensions.FileSystemGlobbing
+{
+    /// <summary>
+    /// Avoids using disk for uses like Pattern Matching.
+    /// </summary>
+    public class InMemoryDirectoryInfo : DirectoryInfoBase
+    {
+        private static readonly char[] DirectorySeparators = new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
+        private readonly IEnumerable<string> _files;
+
+        /// <summary>
+        /// Creates a new InMemoryDirectoryInfo with the root directory and files given.
+        /// </summary>
+        /// <param name="rootDir">The root directory that this FileSystem will use.</param>
+        /// <param name="files">Collection of file names. If relative paths <paramref name="rootDir"/> will be prepended to the paths.</param>
+        public InMemoryDirectoryInfo(string rootDir, IEnumerable<string> files)
+            : this(rootDir, files, false)
+        {
+        }
+
+        private InMemoryDirectoryInfo(string rootDir, IEnumerable<string> files, bool normalized)
+        {
+            if (string.IsNullOrEmpty(rootDir))
+            {
+                throw new ArgumentNullException(nameof(rootDir));
+            }
+
+            if (files == null)
+            {
+                files = new List<string>();
+            }
+
+            Name = Path.GetFileName(rootDir);
+            if (normalized)
+            {
+                _files = files;
+                FullName = rootDir;
+            }
+            else
+            {
+                var fileList = new List<string>(files.Count());
+
+                // normalize
+                foreach (var file in files)
+                {
+                    fileList.Add(Path.GetFullPath(file.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)));
+                }
+
+                _files = fileList;
+
+                FullName = Path.GetFullPath(rootDir.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
+            }
+        }
+
+        /// <inheritdoc />
+        public override string FullName { get; }
+
+        /// <inheritdoc />
+        public override string Name { get; }
+
+        /// <inheritdoc />
+        public override DirectoryInfoBase ParentDirectory
+        {
+            get
+            {
+                return new InMemoryDirectoryInfo(Path.GetDirectoryName(FullName), _files, true);
+            }
+        }
+
+        /// <inheritdoc />
+        public override IEnumerable<FileSystemInfoBase> EnumerateFileSystemInfos()
+        {
+            var dict = new Dictionary<string, List<string>>();
+            foreach (var file in _files)
+            {
+                if (!IsRootDirectory(FullName, file))
+                {
+                    continue;
+                }
+
+                var endPath = file.Length;
+                var beginSegment = FullName.Length + 1;
+                var endSegment = file.IndexOfAny(DirectorySeparators, beginSegment, endPath - beginSegment);
+
+                if (endSegment == -1)
+                {
+                    yield return new InMemoryFileInfo(file, this);
+                }
+                else
+                {
+                    var name = file.Substring(0, endSegment);
+                    List<string> list;
+                    if (!dict.TryGetValue(name, out list))
+                    {
+                        dict[name] = new List<string> { file };
+                    }
+                    else
+                    {
+                        list.Add(file);
+                    }
+                }
+            }
+
+            foreach (var item in dict)
+            {
+                yield return new InMemoryDirectoryInfo(item.Key, item.Value, true);
+            }
+        }
+
+        private bool IsRootDirectory(string rootDir, string filePath)
+        {
+            if (!filePath.StartsWith(rootDir, StringComparison.Ordinal) ||
+                filePath.IndexOf(Path.DirectorySeparatorChar, rootDir.Length) != rootDir.Length)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <inheritdoc />
+        public override DirectoryInfoBase GetDirectory(string path)
+        {
+            if (string.Equals(path, "..", StringComparison.Ordinal))
+            {
+                return new InMemoryDirectoryInfo(Path.Combine(FullName, path), _files, true);
+            }
+            else
+            {
+                var normPath = Path.GetFullPath(path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
+                return new InMemoryDirectoryInfo(normPath, _files, true);
+            }
+        }
+
+        /// <summary>
+        /// Returns an instance of <see cref="FileInfoBase"/> that matches the <paramref name="path"/> given.
+        /// </summary>
+        /// <param name="path">The filename.</param>
+        /// <returns>Instance of <see cref="FileInfoBase"/> if the file exists, null otherwise.</returns>
+        public override FileInfoBase GetFile(string path)
+        {
+            var normPath = Path.GetFullPath(path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
+            foreach (var file in _files)
+            {
+                if (string.Equals(file, normPath))
+                {
+                    return new InMemoryFileInfo(file, this);
+                }
+            }
+
+            return null;
+        }
+    }
+}
--- a/src/Microsoft.Extensions.FileSystemGlobbing/MatcherExtensions.cs
+++ b/src/Microsoft.Extensions.FileSystemGlobbing/MatcherExtensions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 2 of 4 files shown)


src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,134 @@
+﻿using System;
```
pakrym: (c)

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.FileSystemGlobbing.Abstractions
+{
+    public class InMemoryDirectoryInfo : DirectoryInfoBase
```
pakrym: Docs

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,134 @@
...
+                    _files[i] = Path.GetFullPath(_files[i].Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
+                }
+            }
+
+            FullName = Path.GetFullPath(rootDir.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
+            Name = Path.GetFileName(rootDir);
+        }
+
+        // This is to avoid the overhead of Path.GetFullPath from the public constructor
+        private InMemoryDirectoryInfo(string rootDir, List<string> files, bool normalized)
```
pakrym: Check normalized argument and do normalization logic in this constructor, call passing false from default one.

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryFileInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class InMemoryFileInfo : FileInfoBase
+    {
+        private InMemoryDirectoryInfo _parent;
+        public InMemoryFileInfo(string file, InMemoryDirectoryInfo parent)
+        {
+            FullName = file;
+            Name = Path.GetFileName(file);
+            _parent = parent;
+        }
+        public override string FullName { get; }
```
pakrym: nit: newline

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryFileInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.FileSystemGlobbing.Abstractions
+{
+    public class InMemoryFileInfo : FileInfoBase
+    {
+        private InMemoryDirectoryInfo _parent;
```
pakrym: nit: newline

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.FileSystemGlobbing.Abstractions
+{
+    public class InMemoryDirectoryInfo : DirectoryInfoBase
+    {
+        private static readonly char[] DirectorySeparators = new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
+        private readonly List<string> _files;
+
+        public InMemoryDirectoryInfo(string rootDir, List<string> files)
```
davidfowl: `IEnumerable<string>`

pranavkm: Or `IReadOnlyList<>` if you'd like to index into it for some reason.

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryFileInfo.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+
+namespace Microsoft.Extensions.FileSystemGlobbing.Abstractions
+{
+    public class InMemoryFileInfo : FileInfoBase
```
MartinJohns: Shouldn't this be private? Or at least documented when it's public?

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// </summary>
+        /// <param name="rootDir">The root directory that this FileSystem will use.</param>
+        /// <param name="files">Collection of file names.</param>
+        public InMemoryDirectoryInfo(string rootDir, IEnumerable<string> files)
+            : this(rootDir, files, false)
+        {
+        }
+
+        private InMemoryDirectoryInfo(string rootDir, IEnumerable<string> files, bool normalized)
+        {
```
moozzyk: check if rootDir is not null

test/Microsoft.Extensions.FileSystemGlobbing.Tests/FunctionalTests.cs
```diff
@@ -394,10 +427,36 @@ public void MultipleSubDirsAfterFirstWildcardMatch_HasCorrectStem()
...
-            context.CreateFiles(
+            var matcher = new Matcher();
+            matcher.AddInclude("compiler/**/*.cs");
+
+            var files = GetFileList();
+            var directoryPath = "src/project";
+            var results = matcher.Execute(new InMemoryDirectoryInfo(directoryPath, files));
+
+            var actual = results.Files.Select(match => match.Stem);
+            var expected = new string[] {
```
pranavkm: Brace on the next line.

test/Microsoft.Extensions.FileSystemGlobbing.Tests/FunctionalTests.cs
```diff
@@ -370,6 +371,38 @@ public void StemCorrectWithDifferentWildCards()
         }
 
         [Fact]
+        public void StemCorrectWithDifferentWildCards_WithInMemory()
+        {
+            var matcher = new Matcher();
+            matcher.AddInclude("sub/*.cs");
+            matcher.AddInclude("**/*.cs");
```
moozzyk: how about just "*.cs"?

BrennanConroy: `*.cs` just does top level, doesn't go into sub directories.

pranavkm:
Should we have a extension API around this feature? 

Instead of `var results = matcher.Execute(new InMemoryDirectoryInfo(directoryPath, files));`, could we have something like `matcher.Match(files);` \ `matcher.Match(pattern, files)`;? 

Also, what does the base dir do in this case? Are the paths absolute and not root relative?

BrennanConroy:
I would be ok with `matcher.Match(files)` and `matcher.Match(root, files)`. Doing `(pattern, files)` is odd since you set patterns using `AddExcludePatterns` and `AddIncludePatterns`

pranavkm:
Sure. The last one was closer to `Regex.Match` - https://msdn.microsoft.com/en-us/library/0z2heewz(v=vs.110).aspx. We can always add it it it's the most common way this gets used.

test/Microsoft.Extensions.FileSystemGlobbing.Tests/FunctionalTests.cs
```diff
@@ -382,7 +383,8 @@ public void StemCorrectWithDifferentWildCards_WithInMemory()
             var results = matcher.Execute(new InMemoryDirectoryInfo(directoryPath, files));
```
pranavkm: Do you want to use the new extension methods here instead? Just so that the new code path is also exercised?

pranavkm:
:shipit: from me. Have @pakrym \ @moozzyk \ @davidfowl sign off on this before it's checked in.

src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,176 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Extensions.FileSystemGlobbing.Abstractions
```
davidfowl: Why is this in abstractions and not a private implementation of the globbing package?

src/Microsoft.Extensions.FileSystemGlobbing/Internal/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Enumerates the files and directories in the root directory.
+        /// </summary>
+        /// <returns>The list of directories and files in the root directory.</returns>
+        public override IEnumerable<FileSystemInfoBase> EnumerateFileSystemInfos()
+        {
+            var dict = new Dictionary<string, List<string>>();
```
victorhurdugaci: From the code it seems that the results of this will not change between runs. Should you cache the result then?

src/Microsoft.Extensions.FileSystemGlobbing/MatcherExtensions.cs
```diff
@@ -40,5 +40,28 @@ public static IEnumerable<string> GetResultsInFullPath(this Matcher matcher, str
...
             return result;
         }
+
+        /// <summary>
+        /// Matches the files passed in with the patterns in the matcher without going to disk.
+        /// </summary>
+        /// <param name="matcher">The matcher that holds the patterns and pattern matching type.</param>
+        /// <param name="files">The files to run the matcher against.</param>
+        /// <returns>The match results.</returns>
+        public static PatternMatchingResult Match(this Matcher matcher, IEnumerable<string> files)
```
victorhurdugaci: Can you add another overload with a single file? That's the more common scenario

src/Microsoft.Extensions.FileSystemGlobbing/MatcherExtensions.cs
```diff
@@ -40,5 +40,28 @@ public static IEnumerable<string> GetResultsInFullPath(this Matcher matcher, str
...
+        /// <summary>
+        /// Matches the files passed in with the patterns in the matcher without going to disk.
+        /// </summary>
+        /// <param name="matcher">The matcher that holds the patterns and pattern matching type.</param>
+        /// <param name="rootDir">The root directory for the matcher to match the files from.</param>
+        /// <param name="files">The files to run the matcher against.</param>
+        /// <returns>The match results.</returns>
+        public static PatternMatchingResult Match(this Matcher matcher, string rootDir, IEnumerable<string> files)
+        {
+            return matcher.Execute(new InMemoryDirectoryInfo(rootDir, files));
```
victorhurdugaci: Null check for `matcher`?

src/Microsoft.Extensions.FileSystemGlobbing/Internal/InMemoryDirectoryInfo.cs
BrennanConroy: This shouldn't be run more than once per instance if the matcher is doing its job correctly.

src/Microsoft.Extensions.FileSystemGlobbing/Internal/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,177 @@
...
+    internal class InMemoryDirectoryInfo : DirectoryInfoBase
+    {
+        private static readonly char[] DirectorySeparators = new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
+        private readonly IEnumerable<string> _files;
+
+        /// <summary>
+        /// Creates a new InMemoryDirectoryInfo with the root directory and files given.
+        /// </summary>
+        /// <param name="rootDir">The root directory that this FileSystem will use.</param>
+        /// <param name="files">Collection of file names.</param>
```
victorhurdugaci: Is it file names or actually file names + relative paths?

src/Microsoft.Extensions.FileSystemGlobbing/MatcherExtensions.cs
victorhurdugaci: Ugh... does it construct the matcher all the time? Running this twice will do a lot of redundant work

victorhurdugaci:
There should also be an API that simply returns the boolean result of the check. All I want is to check that path matches a pattern. I don't care about `MatchResult` or how this thing works internally.

victorhurdugaci:
:watch:

src/Microsoft.Extensions.FileSystemGlobbing/MatcherExtensions.cs
BrennanConroy: Yeah it will :(
Any suggested workaround besides exposing the InMemoryDirectoryInfo

pranavkm: Should be fine to have the `InMemoryDirectoryInfo` available, but recommend using this for the most common case.

BrennanConroy:
🆙 📅

BrennanConroy:
🆙 📅

src/Microsoft.Extensions.FileSystemGlobbing/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                return new InMemoryDirectoryInfo(Path.GetDirectoryName(FullName), _files, true);
+            }
+        }
+
+        /// <summary>
+        /// Enumerates the files and directories in the root directory.
+        /// </summary>
+        /// <returns>The list of directories and files in the root directory.</returns>
+        public override IEnumerable<FileSystemInfoBase> EnumerateFileSystemInfos()
```
natemcmaster: You can use `/// <inheritdoc />` when the [xmldoc on the base class](https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.Extensions.FileSystemGlobbing/Abstractions/DirectoryInfoBase.cs#L13-L17) is sufficient to describe this overload.

natemcmaster:
:shipit: on implementation from me. Previous reviewers: are your concerns resolved?

src/Microsoft.Extensions.FileSystemGlobbing/InMemoryDirectoryInfo.cs
```diff
@@ -0,0 +1,164 @@
...
+            foreach (var item in dict)
+            {
+                yield return new InMemoryDirectoryInfo(item.Key, item.Value, true);
+            }
+        }
+
+        private bool IsRootDirectory(string rootDir, string filePath)
+        {
+            if (!filePath.StartsWith(rootDir, StringComparison.Ordinal)
+                || filePath.IndexOf(Path.DirectorySeparatorChar, rootDir.Length) != rootDir.Length)
```
victorhurdugaci: nit: `||` on the previous line, not on new line

victorhurdugaci:
:shipit:

src/Microsoft.Extensions.FileSystemGlobbing/InMemoryDirectoryInfo.cs
MartinJohns: Any reason why? IMO I consider it on the new line to be much more readable and easier to understand. I can immediately see that these two conditions are combined with a logical OR, instead of having to go to the earlier line and look all the way to the right to find the end (and the length various depending on the statement).

pranavkm: @MartinJohns it's the formatting we use elsewhere in our codebase.

PR #235, opened 2016-09-20
Modify PhysicalFileWatcher.Watch to work for new files being added

### LINKED ISSUE #233: Modify PhysicalFileWatcher.Watch to work for new files being added
Currently the PhysicalFileSystem.Watch(“*.cshtml”) does not trigger change events for new files added after the fact. This was done as a low cost solution when adding support for polling. 

We'll modify this behavior to allow both FileSystemWatcher backed events and polling to work for globbing patterns.

pranavkm:
Fixes #233

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Abstractions/project.json
+++ b/src/Microsoft.Extensions.FileProviders.Abstractions/project.json
@@ -3,7 +3,7 @@
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
-    "xmlDoc": false
+    "xmlDoc": true
   },
   "description": "Abstractions of files and directories.\r\nCommonly used types:\r\nMicrosoft.Extensions.FileProviders.IDirectoryContents\r\nMicrosoft.Extensions.FileProviders.IFileInfo\r\nMicrosoft.Extensions.FileProviders.IFileProvider",
   "packOptions": {
--- a/src/Microsoft.Extensions.FileProviders.Abstractions/project.json
+++ b/src/Microsoft.Extensions.FileProviders.Abstractions/project.json
@@ -3,7 +3,7 @@
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
-    "xmlDoc": true
+    "xmlDoc": false
   },
   "description": "Abstractions of files and directories.\r\nCommonly used types:\r\nMicrosoft.Extensions.FileProviders.IDirectoryContents\r\nMicrosoft.Extensions.FileProviders.IFileInfo\r\nMicrosoft.Extensions.FileProviders.IFileProvider",
   "packOptions": {
--- a/src/Microsoft.Extensions.FileProviders.Physical/Internal/IClock.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/Internal/IClock.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.FileProviders.Physical.Internal
+{
+    public interface IClock
+    {
+        DateTime UtcNow { get; }
+    }
+}
--- a/src/Microsoft.Extensions.FileProviders.Physical/Internal/IncrementalHash.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/Internal/IncrementalHash.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if NET451
+using System;
+using System.Security.Cryptography;
+
+namespace Microsoft.Extensions.FileProviders.Physical.Internal
+{
+    internal class IncrementalHash : IDisposable
+    {
+        private static readonly byte[] EmptyArray = new byte[0];
+        private readonly SHA256 _sha256 = CreateSHA256();
+        private bool _initialized;
+
+        public void AppendData(byte[] data, int offset, int count)
+        {
+            EnsureInitialized();
+            _sha256.TransformBlock(data, offset, count, outputBuffer: null, outputOffset: 0);
+        }
+
+        public byte[] GetHashAndReset()
+        {
+            _sha256.TransformFinalBlock(EmptyArray, 0, 0);
+            return _sha256.Hash;
+        }
+
+        public void Dispose()
+        {
+            _sha256.Dispose();
+        }
+
+        private void EnsureInitialized()
+        {
+            if (!_initialized)
+            {
+                _sha256.Initialize();
+                _initialized = true;
+            }
+        }
+
+        private static SHA256 CreateSHA256()
+        {
+            SHA256 sha256;
+            try
+            {
+                sha256 = SHA256.Create();
+            }
+            catch (System.Reflection.TargetInvocationException)
+            {
+                // SHA256.Create is documented to throw this exception on FIPS compliant machines.
+                // See: https://msdn.microsoft.com/en-us/library/z08hz7ad%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396
+                // Fallback to a FIPS compliant SHA256 algorithm.
+                sha256 = new SHA256CryptoServiceProvider();
+            }
+
+            return sha256;
+        }
+    }
+}
+#endif
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
@@ -3,13 +3,11 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.IO;
 using System.Security;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.FileSystemGlobbing;
-using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Extensions.FileProviders.Physical
@@ -24,7 +22,9 @@ namespace Microsoft.Extensions.FileProviders.Physical
     /// </summary>
     public class PhysicalFilesWatcher : IDisposable
     {
-        private readonly ConcurrentDictionary<string, ChangeTokenInfo> _matchInfoCache =
+        private readonly ConcurrentDictionary<string, ChangeTokenInfo> _filePathTokenLookup =
+            new ConcurrentDictionary<string, ChangeTokenInfo>(StringComparer.OrdinalIgnoreCase);
+        private readonly ConcurrentDictionary<string, ChangeTokenInfo> _wildcardTokenLookup =
             new ConcurrentDictionary<string, ChangeTokenInfo>(StringComparer.OrdinalIgnoreCase);
 
         private readonly FileSystemWatcher _fileWatcher;
@@ -82,38 +82,37 @@ namespace Microsoft.Extensions.FileProviders.Physical
 
             filter = NormalizePath(filter);
 
+            var changeToken = GetOrAddChangeToken(filter);
+            TryEnableFileSystemWatcher();
+
+            return changeToken;
+        }
+
+        private IChangeToken GetOrAddChangeToken(string pattern)
+        {
             IChangeToken changeToken;
-            var isWildCard = filter.IndexOf('*') != -1;
-            if (isWildCard || IsDirectoryPath(filter))
+            var isWildCard = pattern.IndexOf('*') != -1;
+            if (isWildCard || IsDirectoryPath(pattern))
             {
-                changeToken = ResolveFileTokensForGlobbingPattern(filter);
+                changeToken = GetOrAddWildcardChangeToken(pattern);
             }
             else
             {
-                changeToken = GetOrAddChangeToken(filter);
-            }
-
-            lock (_lockObject)
-            {
-                if (_matchInfoCache.Count > 0 && !_fileWatcher.EnableRaisingEvents)
-                {
-                    // Perf: Turn on the file monitoring if there is something to monitor.
-                    _fileWatcher.EnableRaisingEvents = true;
-                }
+                changeToken = GetOrAddFilePathChangeToken(pattern);
             }
 
             return changeToken;
         }
 
-        private IChangeToken GetOrAddChangeToken(string filePath)
+        private IChangeToken GetOrAddFilePathChangeToken(string filePath)
         {
             ChangeTokenInfo tokenInfo;
-            if (!_matchInfoCache.TryGetValue(filePath, out tokenInfo))
+            if (!_filePathTokenLookup.TryGetValue(filePath, out tokenInfo))
             {
                 var cancellationTokenSource = new CancellationTokenSource();
                 var cancellationChangeToken = new CancellationChangeToken(cancellationTokenSource.Token);
                 tokenInfo = new ChangeTokenInfo(cancellationTokenSource, cancellationChangeToken);
-                tokenInfo = _matchInfoCache.GetOrAdd(filePath, tokenInfo);
+                tokenInfo = _filePathTokenLookup.GetOrAdd(filePath, tokenInfo);
             }
 
             IChangeToken changeToken = tokenInfo.ChangeToken;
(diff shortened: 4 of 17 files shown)


pranavkm:
cc @BrennanConroy \ @rynowak

src/Microsoft.Extensions.FileProviders.Physical/Internal/IncrementalHash.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if NET451
+using System;
+using System.Security.Cryptography;
+
+namespace Microsoft.Extensions.FileProviders.Physical.Internal
+{
+    internal class IncrementalHash : IDisposable
```
pranavkm: Based on https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/IncrementalHash.net46.cs

src/Microsoft.Extensions.FileProviders.Physical/Internal/IClock.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.FileProviders.Physical.Internal
+{
+    public interface IClock
```
natemcmaster: Can we make all the clock abstractions should be true `internal`?

src/Microsoft.Extensions.FileProviders.Abstractions/project.json
```diff
@@ -3,7 +3,7 @@
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
-    "xmlDoc": true
+    "xmlDoc": false
```
natemcmaster: Sadness...after all that effort to add xmldocs...

src/Microsoft.Extensions.FileSystemGlobbing/PatternMatchingResult.cs
```diff
@@ -15,13 +16,30 @@ public class PatternMatchingResult
         /// </summary>
         /// <param name="files">A collection of <see cref="FilePatternMatch" /></param>
         public PatternMatchingResult(IEnumerable<FilePatternMatch> files)
+            : this(files, hasMatches: files.Any())
```
natemcmaster: Can we lazy-evaluate this?

src/Microsoft.Extensions.FileSystemGlobbing/PatternMatchingResult.cs
```diff
@@ -15,13 +16,30 @@ public class PatternMatchingResult
...
+
+        /// <summary>
         /// A collection of <see cref="FilePatternMatch" />
         /// </summary>
         public IEnumerable<FilePatternMatch> Files { get; set; }
+
+        /// <summary>
+        /// Gets a value that determines if this instance of <see cref="PatternMatchingResult"/> has any matches.
+        /// </summary>
+        public bool HasMatches { get; }
```
natemcmaster: When will this ever differ from `Files.Any()`?

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Security.Cryptography;
+using System.Text;
+using Microsoft.Extensions.FileProviders.Physical.Internal;
+using Microsoft.Extensions.FileSystemGlobbing;
+using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Physical
+{
+    public class PollingWildCardChangeToken : IChangeToken
```
natemcmaster: xmldoc.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+            string root,
+            string pattern)
+            : this(
+                new DirectoryInfoWrapper(new DirectoryInfo(root)),
+                pattern,
+                Internal.Clock.Instance)
+        {
+        }
+
+        public PollingWildCardChangeToken(
```
natemcmaster: `internal` constructor

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+                        // _lastScanTimeUtc is the greatest timestamp that any last writes could have been.
+                        // If a file has a newer timestamp than this value, it must've changed.
+                        hasChanges = true;
+                    }
+
+                    ComputeHash(sha256, file.Path);
+                }
+
+                var currentHash = sha256.GetHashAndReset();
+                hasChanges |= !ArrayEquals(_previousHash, currentHash);
```
natemcmaster: Enumerable.SequenceEqual?

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (_fileNameBuffer == null || byteCount > _fileNameBuffer.Length)
+            {
+                _fileNameBuffer = new byte[Math.Max(byteCount, 256)];
+            }
+
+            var length = Encoding.UTF8.GetBytes(path, 0, path.Length, _fileNameBuffer, 0);
+            sha256.AppendData(_fileNameBuffer, 0, length);
+        }
+
+        /// <inheritdoc />
```
natemcmaster: Add explanation for that this method doesn't actually do anything.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+#endif
+            {
+                foreach (var file in files)
+                {
+                    if (!hasChanges &&
+                        _lastScanTimeUtc < GetLastWriteUtc(file.Path))
+                    {
+                        // _lastScanTimeUtc is the greatest timestamp that any last writes could have been.
+                        // If a file has a newer timestamp than this value, it must've changed.
+                        hasChanges = true;
```
natemcmaster: This can short circuit and stop computing the hash on files.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private void ComputeHash(IncrementalHash sha256, string path)
+        {
+            var byteCount = Encoding.UTF8.GetByteCount(path);
```
natemcmaster: Can we just leave this in Unicode?

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
```diff
@@ -230,20 +240,59 @@ private void ReportChangeForMatchedEntries(string path)
...
+            if (matched)
+            {
+                TryDisableFileSystemWatcher();
+            }
+        }
 
-                if (_matchInfoCache.Count == 0)
+        private void TryDisableFileSystemWatcher()
+        {
+            if (_filePathTokenLookup.IsEmpty && _wildcardTokenLookup.IsEmpty)
```
BrennanConroy: @davidfowl told me not to use the if-lock-if pattern because it can break on hardware such as ARM

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            _matcher = new Matcher(StringComparison.OrdinalIgnoreCase);
+            _matcher.AddInclude(pattern);
+            CalculateChanges();
+        }
+
+        /// <inheritdoc />
+        public bool ActiveChangeCallbacks => false;
+
+        public IClock Clock { get; }
```
BrennanConroy: IClock is in internal namespace...

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
BrennanConroy: It looks fine to me, the hashes for every file are computed and the GetLastWriteUtc check is skipped if any files have changed. Unless you meant we could short circuit here if hasChanges is never true...

src/Microsoft.Extensions.FileProviders.Physical/project.json
```diff
@@ -38,10 +41,12 @@
     "netstandard1.3": {
       "dependencies": {
         "System.Collections.Concurrent": "4.0.12-*",
+        "System.Diagnostics.Debug": "4.0.11",
         "System.IO.FileSystem.Watcher": "4.0.0-*",
         "System.IO.FileSystem": "4.0.1-*",
         "System.Runtime.Extensions": "4.1.0-*",
-        "System.Threading": "4.0.11-*"
+        "System.Threading": "4.0.11-*",
+        "System.Security.Cryptography.Algorithms": "4.2.0"
```
BrennanConroy: nit: Sort

src/Microsoft.Extensions.FileSystemGlobbing/PatternMatchingResult.cs
pranavkm: We use the new constructor which accepts the `hasMatches` values as an argument. The callsite does a `List<>.Count > 0` which avoids the `.Any()`. The other ctor isn't used in our code, but we can't change because that'd be breaking. Essentially it's never different.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
pranavkm: Unfortunately not. The hash is the only way to know if files were removed (since that wouldn't change the last changed timestamp).

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
pranavkm: Boo Linq.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
pranavkm: Sure. It made the code messier anyway.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
pranavkm: We could consider commonizing this with the one in Caching and Kestrel - https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/ISystemClock.cs. Security has this type too but it's public API: https://github.com/aspnet/Security/blob/22d133cefa20926b25eeb82779343b23a221117f/src/Microsoft.AspNetCore.Authentication/ISystemClock.cs

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
pranavkm: Is it cheaper to go from string -> unicode bytes?

src/Microsoft.Extensions.FileSystemGlobbing/PatternMatchingResult.cs
pranavkm: None of our code calls this ctor anymore. Would be unnecessary to make  the code messier when we don't have a good reason to.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+                return _changed;
+            }
+        }
+
+        private bool CalculateChanges()
+        {
+            var result = _matcher.Execute(_directoryInfo);
+            var hasChanges = false;
+
+            var files = result.Files.OrderBy(f => f.Path, StringComparer.OrdinalIgnoreCase);
```
pranavkm: It'd be really nice if we can guarantee ordering from the matcher. It relies on filesystem enumeration which does return results in alphabetic order to do this on all platforms but it's not part of the contract.

src/Microsoft.Extensions.FileProviders.Abstractions/project.json
pranavkm: I should probably revert this file. Perhaps making things `internal` in the other project would let me revert it in the second project too.

natemcmaster: To be clear, this suppresses xml doc generation completely. If you just want to disable the warning, use "nowarn"

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
natemcmaster: I'm not sure its a useful abstraction. IIUC we only use this for testing.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
natemcmaster: Once you've identified `hasChanges = true`, the hash will never be used again. There's no need to compute it.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            return hasChanges;
+        }
+
+        protected virtual DateTime GetLastWriteUtc(string path)
+        {
+            return File.GetLastWriteTimeUtc(Path.Combine(_directoryInfo.FullName, path));
+        }
+
+        private bool ArrayEquals(byte[] _previousHash, byte[] currentHash)
```
natemcmaster: Nit: don't use underscore in name

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
natemcmaster: Although it basically reimplements a BCL method, your implementation allocates less... so keep it :)

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            if (_previousHash == null)
+            {
+                // First run
+                return true;
+            }
+
+            Debug.Assert(_previousHash.Length == currentHash.Length);
```
natemcmaster: Return false?

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
natemcmaster: My original thinking was .NET stores strings as UTF-16 (Unicode) so you would think its cheaper. But, it looks like the [implementation](http://referencesource.microsoft.com/#mscorlib/system/text/unicodeencoding.cs,2b03a0eac68f75f2) of `UnicodeEncoding.GetBytes` isn't much better that `UTF8Encoding.GetBytes`. It's be fastest to do this in unsafe code and use the string's raw memory (assuming the sha hash doesn't manipulate the original memory).

src/Microsoft.Extensions.FileSystemGlobbing/PatternMatchingResult.cs
natemcmaster: In that case, we should ensure the symmetry of `HasChanges` and `Files.Any()` by removing the constructor that takes `bool hasMatches`. We could just initialize this value by calling Files.Any() once in the original ctor

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
natemcmaster: This can be case sensitive sort (default comparer). If case changed, the hash will be different anyways and order won't matter.

pranavkm:
Bump

halter73:
If we don't "commonize" this with Kestrel and Caching now, is there anything preventing us from doing so later?

pranavkm:
@halter73 I did not know we had file system watching in the other two repos. Why does Kestrel watch files?

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,171 @@
...
+        // Internal for unit testing.
+        internal PollingWildCardChangeToken(
+            DirectoryInfoBase directoryInfo,
+            string pattern,
+            IClock clock)
+        {
+            _directoryInfo = directoryInfo;
+            Clock = clock;
+
+            _matcher = new Matcher(StringComparison.OrdinalIgnoreCase);
```
BrennanConroy: Anyway to pass this in from https://github.com/aspnet/FileSystem/pull/235/files#diff-8bb8e51541651d6e13df23d9a7311bb1R141

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -0,0 +1,171 @@
...
+            get
+            {
+                if (_changed)
+                {
+                    return _changed;
+                }
+
+                if (Clock.UtcNow - _lastScanTimeUtc >= PollingInterval)
+                {
+                    lock (_enumerationLock)
```
BrennanConroy: This is the only place this lock object is used

src/Microsoft.Extensions.FileProviders.Physical/project.json
```diff
@@ -38,9 +38,11 @@
     "netstandard1.3": {
       "dependencies": {
         "System.Collections.Concurrent": "4.0.12-*",
+        "System.Diagnostics.Debug": "4.0.11",
```
BrennanConroy: -*?

halter73:
@pranavkm I should have been more clear. Kestrel doesn't watch files. I was referring to https://github.com/aspnet/Common/issues/151. It might be nice to have for perf/maintenance, but I don't  think it's that big of a deal since the needs are different enough.

pranavkm:
@halter73 apparently nobody cared for the type given that it's pretty small.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
pranavkm: Might result in concurrency issues (since FSW's calls to the `Matcher` might get called in the background while this is running). The type isn't thread safe so it's more work to manage the concurrency.

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
pranavkm: It's basically to prevent concurrent calls to the Matcher

src/Microsoft.Extensions.FileProviders.Physical/project.json
pranavkm: This is going away: https://github.com/aspnet/FileSystem/commit/0bf9850d71ddf91473052ecff9c9087326f01fa8

BrennanConroy:
:shipit:

Issue #224, opened 2016-08-08
Delay enumeration of directory until IDirectoryContents enumerated


davidfowl:
Not sure why we do this inline, but I was profiling the static file middleware and saw enumeration even though it's just checking for directory existence:

https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs#L219-L236

https://github.com/aspnet/StaticFiles/blob/dev/src/Microsoft.AspNetCore.StaticFiles/DefaultFilesMiddleware.cs#L69-L95

Yves57:
I just had a somewhat similar problem in a WebDAV middleware implementation: I just want to check a path validity, and if this path represents a file or a directory.
Calling `GetFileInfo` works for the files but not for the directories. So if it fails, I have to call `GetDirectoryContents` to check if it is a directory. And of course it enumerates all child entries for nothing in my case.

So just for curiosity, why `GetFileInfo` does not return a valid `IFileInfo` for a directory? For performance reasons in case of the path is invalid?

natemcmaster:
Might be room for even more improvement that #237. See https://github.com/aspnet/FileSystem/pull/237#issuecomment-250090925

PR #237, opened 2016-09-26
Defer creation of file info in directory contents enumeration

natemcmaster:
Resolves https://github.com/aspnet/FileSystem/issues/224

cc @Tratcher @troydai

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ project.lock.json
 .build
 .testPublish
 .idea/
+*.nuget.props
+*.nuget.targets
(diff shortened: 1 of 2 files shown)


Tratcher:
:shipit:
Does it concern anyone that the enumeration is now unstable? If you enumerate it multiple times you may now get different results.

natemcmaster:
Hmm, that's a good insight, but it's not something our API every guaranteed. IEnumerable implies lazy-loaded enumeration which might vary from run to run. IMO callers who need to ensure stability over multiple iterations should have always used ToList anyways.

davidfowl:
@natemcmaster why not move the enumeration further into the EnumerableDirectoryContents itself? Basically avoid any work until somebody asks for the enumerator? It's worth measuring the static files middleware to see what the difference is

PR #239, opened 2016-09-29
Modify PollingWildCardChangeToken to include the timestamp of the files

### LINKED ISSUE #238: Modify PollingWildCardChangeToken to include the timestamp of the files along with the names as part of hash check
Consider the following timeline

```
t0 ->  Create a copy of  File.txt to a backup directory . 
t1 -> Modify File.txt with contents different from the original file.
t2 -> Move the file from the backup directory to the original location.
```

The `PollingWildCardChangeToken` only carries the timestamp of the most recently changed last write time window. Consequently this wouldn't work when a file is modified but it's timestamp hasn't moved forward. We could avoid this by additionally storing the timestamp along with the name of the file.

pranavkm:
Fixes #238

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Extensions.FileProviders.Physical
         private readonly Matcher _matcher;
         private bool _changed;
         private DateTime? _lastScanTimeUtc;
-        private byte[] _fileNameBuffer;
+        private byte[] _byteBuffer;
         private byte[] _previousHash;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


pranavkm:
cc @Eilon \ @BrennanConroy

BrennanConroy:
:sheep: :it:

src/Microsoft.Extensions.FileProviders.Physical/PollingWildCardChangeToken.cs
```diff
@@ -153,16 +154,26 @@ private static bool ArrayEquals(byte[] previousHash, byte[] currentHash)
...
-            sha256.AppendData(_fileNameBuffer, 0, length);
+            var length = Encoding.Unicode.GetBytes(value, 0, value.Length, _byteBuffer, 0);
+            sha256.AppendData(_byteBuffer, 0, length);
+
+            Debug.Assert(_byteBuffer.Length > sizeof(long));
+            unsafe
+            {
+                fixed (byte* b = _byteBuffer)
+                {
+                    *((long*)b) = lastChangedUtc.Ticks;
```
Eilon: Do we typically do unsafe code to avoid all the bit masking that would otherwise be required here?

pranavkm: Not sure I fully understand. I'll bother you offline.

Eilon: We talked offline. This is gross but it's ok.

PR #240, opened 2016-10-07
Reduce allocations in PhysicalFileProvider when GetDirectoryContents is called but not enumerated

### LINKED ISSUE #224: Delay enumeration of directory until IDirectoryContents enumerated

natemcmaster:
Resolves #224 

Optimize PhysicalFileProvider for scenario when it is basically being used for Directory.Exists() (cough, looking at you, aspnet/StaticFiles).

Add singleton of NotFoundDirectoryContents.
Delay calling DirectoryInfo.EnumerateFileSystemInfos which does lots of up-front allocation.

**Benchmark: 100 checks for directory that exists but contents not enumerated**
Importantly, this isn't creating an enumerator. We just want to know if the directory exist.
```c#
bool exists;
for (var i = 0; i < 100; i++)
{
    var dirContents = provider.GetDirectoryContents("exists");
    exists = dirContents.Exists;
}
```
_Before:_

image

_After:_
image

 **Benchmark: 100 checks for directory that **does not exists****
```c#
bool exists;
for (var i = 0; i < 100; i++)
{
    var dirContents = provider.GetDirectoryContents("doesnotexist");
    exists = dirContents.Exists;
    var e = dirContents.GetEnumerator();
}
```

_Before:_
image

_After:_
image

cc @davidfowl @Tratcher


natemcmaster:
Typo. This resolves https://github.com/aspnet/FileSystem/issues/224 not 214

src/Microsoft.Extensions.FileProviders.Abstractions/NotFoundDirectoryContents.cs
```diff
@@ -13,6 +13,11 @@ namespace Microsoft.Extensions.FileProviders
     public class NotFoundDirectoryContents : IDirectoryContents
     {
         /// <summary>
+        /// A shared instance of <see cref="NotFoundDirectoryContents"/> 
+        /// </summary>
+        public static NotFoundDirectoryContents Singleton { get; } = new NotFoundDirectoryContents();
```
pranavkm: `Instance`

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -291,5 +272,29 @@ public IChangeToken Watch(string filter)
 
             return _filesWatcher.CreateFileChangeToken(filter);
         }
+
+        private static IDirectoryContents GetEnumerableContents(string fullPath)
+        {
+            return new EnumerableDirectoryContents(() =>
```
pranavkm: Couldn't we pass in the `DirectoryInfo` to `EnumerableDirectoryContents` and have it iterate it lazily (instead of passing in the Func?)

src/Microsoft.Extensions.FileProviders.Sources/EnumerableDirectoryContents.cs
```diff
@@ -22,12 +41,19 @@ public bool Exists
...
 
         IEnumerator IEnumerable.GetEnumerator()
         {
+            EnsureInitialized();
             return _entries.GetEnumerator();
         }
+
+        private void EnsureInitialized()
+        {
+            LazyInitializer.EnsureInitialized(ref _entries, _entriesFactory);
```
pranavkm: Probably don't need any of the behavior of `LazyInitializer`. if (_entries == null) { .. }` should suffice.

src/Microsoft.Extensions.FileProviders.Abstractions/NotFoundDirectoryContents.cs
natemcmaster: We've use `Singleton` in abstractions, e.g. NullChangeToken.Singleton. Seems like they mean the same thing. Is there some nuance I'm missing?

pranavkm: Gotcha. Never mind in that case.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
pranavkm: Could we fork it and make a `PhysicalDirectoryContents`? There's not much to implement in the interface.

natemcmaster: I considered that too and started an implementation with that approach, but ended up with the current implementation. There are probably a dozen ways it could be done, but I landed on this. I couldn't see a strong case for adding a new type such as `PhysicalDirectoryContents`.

src/Microsoft.Extensions.FileProviders.Sources/EnumerableDirectoryContents.cs
natemcmaster: Fair enough. Will update.

natemcmaster:
Ping. @Tratcher can you also review?

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -220,45 +220,26 @@ public IDirectoryContents GetDirectoryContents(string subpath)
...
-                            return new PhysicalDirectoryInfo((DirectoryInfo)info);
-                        }
-                        // shouldn't happen unless BCL introduces new implementation of base type
-                        throw new InvalidOperationException("Unexpected type of FileSystemInfo");
-                    });
-
-                return new EnumerableDirectoryContents(contents);
+                return GetEnumerableContents(fullPath);
             }
             catch (DirectoryNotFoundException)
```
Tratcher: or duplicate it..

natemcmaster:
:up: 📅

PR #250, opened 2016-11-30
Change CompositeFileChangeToken to CompositeChangeToken

jbagga:
Addresses aspnet/Common [issue #163](https://github.com/aspnet/common/issues/163)
PR (https://github.com/aspnet/Common/pull/175)
@pranavkm

PR #251, opened 2016-12-05
Compile shared sources from folder, not package

natemcmaster:
Looking ahead to the MSBuild future. Remove Microsoft.Extensions.FileProviders.Sources package and compile from shared folder.
AFAIK no other aspnet repos use this package show shouldn't need to keep publishing it on our CI feeds.

cref https://github.com/aspnet/Coherence-Signed/issues/379

cc @pranavkm @Eilon


src/Microsoft.Extensions.FileProviders.Abstractions/project.json
```diff
@@ -3,7 +3,8 @@
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
-    "xmlDoc": true
+    "xmlDoc": true,
+    "compile": "../Shared/**/*.cs"
```
pranavkm: lower case "shared"? (Since we have `src`, `tools`, `test` etc)

natemcmaster:
Ok, will wait for #250 and then remove the dead code.

Re: `EmptyDisposable`. Seems like a good API for .NET to add or to move into aspnet/Common. It's not filesystem specific.

test/Microsoft.Extensions.FileProviders.Composite.Tests/project.json
```diff
@@ -1,9 +1,14 @@
...
+    "warningsAsErrors": true,
+    "keyFile": "../../tools/Key.snk",
+    "compile": "../../shared/**/*.cs"
+  },
   "dependencies": {
     "dotnet-test-xunit": "2.2.0-*",
     "Microsoft.AspNetCore.Testing": "1.2.0-*",
     "Microsoft.Extensions.FileProviders.Composite": "1.2.0-*",
-    "Microsoft.Extensions.FileProviders.Sources": "1.2.0-*",
+    "Moq": "4.6.38-alpha",
```
JunTaoLuo: Is there a feature we are using from this alpha version? I see else where in our repos we're using another version: https://github.com/aspnet/BasicMiddleware/blob/ea0042c685ba2228d0d4b85df41fa45a0b360f00/test/Microsoft.AspNetCore.ResponseCompression.Tests/project.json#L15

natemcmaster: 4.6.36 is also an alpha. We aren't uniform on this one though. Other repos use the .38 patch.

natemcmaster:
Yes, Jass is working on removing some of this duplication in #250. This PR is mostly about getting off the "shared" sources project.json model since that package layout will not work in vs2017.

PR #252, opened 2016-12-15
Upgrade to VS 2017

natemcmaster:
Depends on aspnet/KoreBuild https://github.com/aspnet/KoreBuild/pull/145

cc @muratg 
cc @jbagga heads up.


src/Microsoft.Extensions.FileSystemGlobbing/Sources/HashCodeCombiner.cs
```diff
@@ -0,0 +1,86 @@
...
+
+// TODO use the package when https://github.com/dotnet/sdk/issues/495
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.Extensions.Internal
+{
+    internal struct HashCodeCombiner
```
muratg: Was this moved from elsewhere in the repo?  Not sure if this is used for the build...

natemcmaster: This was added from aspnet/Common because we can't take a reference on Microsoft.Extensions.HashCodeCombiner.Sources. cref https://github.com/dotnet/sdk/issues/495

natemcmaster: Tracking its removal: https://github.com/aspnet/Common/issues/180

PR #260, opened 2017-03-09
No mono

davidfowl:


natemcmaster:
:shipit: ...in fact just do this on all our repos. I don't think we use Mono anywhere

PR #274, opened 2017-05-22
Fix: PollingFileChangeToken wasn't watching actual files

BrennanConroy:
cc @JunTaoLuo since you showed interest

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Security;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.FileProviders.Physical.Internal;
 using Microsoft.Extensions.FileSystemGlobbing;
 using Microsoft.Extensions.Primitives;
 
@@ -82,6 +83,12 @@ namespace Microsoft.Extensions.FileProviders.Physical
 
             filter = NormalizePath(filter);
 
+            // Absolute paths and paths traversing above root not permitted.
+            if (Path.IsPathRooted(filter) || PathUtils.PathNavigatesAboveRoot(filter))
+            {
+                return NullChangeToken.Singleton;
+            }
+
             var changeToken = GetOrAddChangeToken(filter);
             TryEnableFileSystemWatcher();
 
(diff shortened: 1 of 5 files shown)


BrennanConroy:
@ryanbrandenburg This fixes the flaky tests on macOS

test/Microsoft.Extensions.FileProviders.Physical.Tests/PhysicalFileProviderTests.cs
```diff
@@ -373,34 +377,11 @@ public void TokenIsSameForSamePath()
...
-                            await Task.Delay(WaitTimeForTokenToFire);
-                            Assert.True(token.HasChanged);
-                        }
-                    }
-                }
-            }
-        }
-
-        [Fact]
-        public async Task WatcherWithPolling_ReturnsTrueForChangedFileWhenQueriedMultipleTimes()
```
BrennanConroy: This was a duplicate test, had 0 code differences

natemcmaster:
Do we have a test that consistently fails before this change and is fixed by this? I'm concerned that we didn't catch this sooner, and hope we can add a test that would help catch regressions in this area.

BrennanConroy:
The 3 affected tests in this PR failed on slower machines occasionally. Catching races is difficult.

natemcmaster:
ok, good

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
```diff
@@ -124,7 +124,7 @@ private IChangeToken GetOrAddFilePathChangeToken(string filePath)
                     new[]
                     {
                         changeToken,
-                        new PollingFileChangeToken(new FileInfo(filePath))
+                        new PollingFileChangeToken(new FileInfo(Path.Combine(_root, filePath)))
```
natemcmaster: Should we first check `Path.IsPathRooted` first? Looking at the code, it seems we never validate `filePath` is relative or absolute.

natemcmaster: This is the code path I found.
PhysicalFilesWatcher.CreateFileChangeToken(string filter) -> GetOrAddChangeToken(filter) -> GetOrAddFilePathChangeToken(filePath)

BrennanConroy: Oh yeah, nice thats a public method

BrennanConroy:
:up: 📅

PR #276, opened 2017-06-07
Target .NET Standard 2.0

natemcmaster:
Part of aspnet/Home#2045.


src/Microsoft.Extensions.FileSystemGlobbing/Util/StringComparisonHelper.cs
```diff
@@ -19,12 +19,10 @@ public static StringComparer GetStringComparer(StringComparison comparisonType)
                     return StringComparer.Ordinal;
                 case StringComparison.OrdinalIgnoreCase:
                     return StringComparer.OrdinalIgnoreCase;
-#if !NETSTANDARD1_3
```
pranavkm: Can you replace this type with `StringComparer.FromComparison` (https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/StringComparer.cs#L76)?

natemcmaster: I wish. It's only part of .NET Core 2.0, not .NET Standard.

PR #278, opened 2017-06-12
Remove optional parameter from DirectoryInfoWrapper ctor

### LINKED ISSUE #229: Make DirectoryInfoWrapper implementation details internal
Microsoft.Extensions.FileSystemGlobbing.Abstractions.DirectoryInfoWrapper has public API that is really only an implementation detail. We could make some of this to internal instead.

Would be a breaking change, so we can't do this until 2.0.

natemcmaster:
Resolves #229. The 'isParentPath' parameter on DirectoryInfoWrapper was only meant to be an internal implementation detail, and should not be exposed publicly.

Technically, this qualifies as a breaking change, which is why we waited until 2.0 to do this. I believe the risk is very low as this is a very obscure API, and the removed parameter was optional to begin with.

cc @muratg


muratg:
Looks good to me. @davidfowl Do you think this would (or should?) be used by anyone else?

natemcmaster:
🔔

natemcmaster:
🔔 again. @davidfowl  or @muratg, okay with this change?

PR #279, opened 2017-06-13
Remove empty helper type from physical file provider

natemcmaster:
One more minor cleanup item.

Microsoft.Extensions.FileProviders.FileSystemInfoHelper was mistakenly made 'public' in 1.0. It is an empty type with no public methods or properties, so although technically breaking, there is no reason anyone should have used this API.

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/FileSystemInfoHelper.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/FileSystemInfoHelper.cs
@@ -5,13 +5,9 @@ using System.IO;
 
 namespace Microsoft.Extensions.FileProviders
 {
-    /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.FileProviders.Physical/Internal/FileSystemInfoHelper.cs
```diff
@@ -3,15 +3,11 @@
 
 using System.IO;
 
-namespace Microsoft.Extensions.FileProviders
+namespace Microsoft.Extensions.FileProviders.Physical.Internal
```
pranavkm: Should be fine to leave it in the root namespace since the type's internal. Alternatively move it to the `.Internal` namespace and leave the type public (I prefer the former).

PR #280, opened 2017-06-20
Add API to allow the physical file provider to show hidden files

### LINKED ISSUE #186: No way to access hidden files
It is good that providers do not enumerate hidden files. It is also impossible to read contents of hidden files in default implementation.

In some cases accessing hidden files is necessary, for example I was trying to access a `.gitignore` file content using `PhysicalFileProvider` with no luck.

Having an optional flag to make hidden files available would be great.

natemcmaster:
Resolves #186 

The default behavior remains consistent: files beginning with a period, Hidden and System files are filtered from the watcher, enumeration, and any other file listing.

The ExclusionFilters enum allows consumers to override this behavior by passing in these flags to one of the new overrides.

cc @dazinator


blowdart:
Not sure about HiddenFileOptions as a name - hidden files has a specific meaning on NTFS, and you're including that, plus system, plus dot files. Maybe ExcludedFiles?

Also is that enum a flag? Shouldn't we then mark it as such, or has that fallen out of favour? Then we'd switch to None / All as names, and remove Show from the others.

natemcmaster:
> Also is that enum a flag?

Yes, it's a flag.

Naming stuff is hard. ExcludedFiles sounds better than HiddenFileOptions. Is this what you had in mind?
```c#
[Flags]
public enum ExcludedFiles
{
   Default = 0xFFFF,
   None = 0,
   StartingWithPeriod = 0x0001,
   Hidden = 0x0002,
   System = 0x0004,
}
```

blowdart:
All rather than default, but yes.

Maybe PeriodPrefixed rather than StartingWithPeriod? Period is also very american, maybe DotPrefixed?

natemcmaster:
DotFiles?

`ExcludedFiles.All` too is ambiguous. Looks like it means "exclude all files regardless of name or attributes", even though it really means "use all filters".

blowdart:
Ah yea, all doesn't make sense. Default is kinda smelly though.

Does Dot* whatever also include directories? If so, DotFiles doesn't work either.

natemcmaster:
Hmm, yup, includes directories. DotPrefixed more correct. That or `ExcludedFiles.FileOrDirectoryStartingWithPeriod_FyiPeriodMeansAscii0x2E`

dazinator:
Naming aside, I like the guts of this.

I still think its weird filtering out dotprefixed files / folders by default - (I think this is special casing for tooling that wants to ignore things like .git folder) but I understand why you want to keep the default behaviour compatible.

Flags for the win.

nil4:
Shouldn't overloads of `GetDirectoryContents` and `Watch` be also added, taking `ExcludedFiles`?

natemcmaster:
> Shouldn't overloads of GetDirectoryContents and Watch be also added, taking ExcludedFiles?

That's a good observation. I originally couldn't come up with a scenario for these overloads that aren't solved by changing options passed to the PhysicalFilesProvider constructor overload. @nil4 do you have scenarios in mind? GetDirectoryContents will be easy to overload. Watch is going to be tricky.

natemcmaster:
Attempt 3 at picking good names
```c#
    [Flags]
    public enum ExcludedFiles
    {
        Sensitive = DotPrefixed | Hidden | System,
        DotPrefixed = 0x0001,
        Hidden = 0x0002,
        System = 0x0004,
        None = 0
    }
```

nil4:
I don't have an actual scenario in mind, it's just that the abstraction seems inconsistent if one can pass `ExcludedFiles` flags to `GetFileInfo` but no other method. Let me ask the question in reverse: assume I'd construct `PhysicalFileProvider` with explicit `ExcludedFiles` flags. What scenario requires passing *different* flags to `GetFileInfo`?

To put it another way, if directory listing and `Watch` coverage is bound by one set of flags, how would my app learn about the existence of a file that *doesn't* match them so that it can call `GetFileInfo` on it?

It seems to me either no overload should take `ExcludedFiles` flag, or all of them do, to make the abstraction self-consistent.

One argument to all methods having overloads would be the following: I'd like to use the default file provider for the static files middleware and not serve sensitive files by default, to keep my app secure. Using the same file provider, I should be able to pass explicit flags in those areas of my app where I need to enumerate the full contents of the file system, bypassing the default filters. 

I agree that `Watch` may be trickier, but wouldn't it have to support filtering by exclude flags anyway, i.e. those passed in the constructor?

nil4:
I just saw your last commit now and it makes my feedback obsolete. Thanks! 👍

natemcmaster:
FYI - I added GetDirectoryContents.

> What scenario requires passing different flags to GetFileInfo?

> I'd like to use the default file provider for the static files middleware and not serve sensitive files by default, to keep my app secure. Using the same file provider, I should be able to pass explicit flags in those areas of my app where I need to enumerate the full contents of the file system, bypassing the default filters.

You basically explained the scenario I'm thinking of. In many places in the code, you don't really want to allocate a new provider just to get a "hidden" file in one part of the code.

natemcmaster:
Working on adding .Watch() too. Not has hard as I thought. Just need to flow the flags to the right place :).

blowdart:
Sensitive works. Because you are a gentle soul :D

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -147,11 +171,24 @@ public IFileInfo GetFileInfo(string subpath)
...
+        /// Enumerate a directory at the given path, if any.
+        /// </summary>
+        /// <param name="subpath">A path under the root directory. Leading slashes are ignored.</param>
+        /// <param name="excludedFiles">Specifies which files will be excluded from the results.</param>
+        /// <returns>
+        /// Contents of the directory. Caller must check <see cref="IDirectoryContents.Exists"/> property. <see cref="NotFoundDirectoryContents" /> if
+        /// <paramref name="subpath" /> is absolute, if the directory does not exist, or <paramref name="subpath" /> has invalid
+        /// characters.
+        /// </returns>
+        public IDirectoryContents GetDirectoryContents(string subpath, ExcludedFiles excludedFiles)
```
Tratcher: Why add this method? It doesn't look like it's used by anything else here, and it's hidden the IFileProvider interface

natemcmaster: Same reason for adding the GetFileInfo overload. I'd like to use the default file provider most of the time, but still be able to bypass its filtering in special cases.

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
```diff
@@ -110,8 +125,17 @@ private bool IsUnderneathRoot(string fullPath)
...
         public IFileInfo GetFileInfo(string subpath)
+            => GetFileInfo(subpath, _excludedFiles);
+
+        /// <summary>
+        /// Locate a file at the given path by directly mapping path segments to physical directories.
+        /// </summary>
+        /// <param name="subpath">A path under the root directory</param>
+        /// <param name="excludedFiles">Overrides default file exclusion behavior.</param>
+        /// <returns>The file information. Caller must check <see cref="IFileInfo.Exists"/> property. </returns>
+        public IFileInfo GetFileInfo(string subpath, ExcludedFiles excludedFiles)
```
pranavkm: This feels weird - why have the options at two different levels? It doesn't seem intuitive, for instance, if I specified `ExcludedFiles.Hidden` in the ctor, and `ExcludedFiles.System` here, I'd sort of expect it to filter both, not just the latter.

test/Microsoft.Extensions.FileProviders.Physical.Tests/ExcludedFilesTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        private readonly DisposableFileSystem _fileSystem = new DisposableFileSystem();
+
+        public void Dispose()
+        {
+            _fileSystem.Dispose();
+        }
+
+        [Theory]
+        [MemberData(nameof(Combinations))]
+        public void FiltersExcludedFiles(string filename, FileAttributes attributes, ExcludedFiles excludedFiles, bool excluded)
```
pranavkm: Make these two separate tests. One for exclusion and one for inclusion.

build/dependencies.props
```diff
@@ -6,6 +6,6 @@
     <RuntimeFrameworkVersion Condition="'$(TargetFramework)'=='netcoreapp2.0'">2.0.0-*</RuntimeFrameworkVersion>
     <TestSdkVersion>15.3.0-*</TestSdkVersion>
     <XunitVersion>2.3.0-beta2-*</XunitVersion>
-    <MoqVersion>4.7.1</MoqVersion>
+    <MoqVersion>4.7.58</MoqVersion>
```
pranavkm: Make this a separate PR. Completely unrelated to this change.

test/Microsoft.Extensions.FileProviders.Physical.Tests/ExcludedFilesTests.cs
natemcmaster: Not sure I understand the request. Am I missing scenarios? Or is this just a coding preference?

build/dependencies.props
natemcmaster: Seems like overkill. We're already in agreement that we want to upgrade. Ryan just hasn't gotten to this repos yet.  https://github.com/aspnet/Coherence-Signed/issues/583.

src/Microsoft.Extensions.FileProviders.Physical/ExcludedFiles.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.IO;
+
+namespace Microsoft.Extensions.FileProviders.Physical
+{
+    /// <summary>
+    /// Specifies while files are excluded by the physical file provider.
+    /// </summary>
+    [Flags]
+    public enum ExcludedFiles
```
blowdart: Ah yes, this still says files when it could also be directories.

src/Microsoft.Extensions.FileProviders.Physical/Internal/FileSystemInfoHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <param name="fileSystemInfo">The file or directory info.</param>
+        /// <param name="excluded">The excluded flag.</param>
+        /// <returns>True when the file is excluded.</returns>
+        public static bool IsExcluded(FileSystemInfo fileSystemInfo, ExcludedFiles excluded)
+        {
+            if (excluded == ExcludedFiles.None)
+            {
+                return false;
+            }
+            else if (fileSystemInfo.Name.StartsWith(".") && (excluded & ExcludedFiles.DotPrefixed) != 0)
```
blowdart: No non ordinal string comparisons! Security!

src/Microsoft.Extensions.FileProviders.Physical/ExcludedFiles.cs
natemcmaster: `ExclusionFilters`?

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs
natemcmaster: Consider another scenario. The ctor was set to `Hidden | DotPrefixed | System`. If I call `.GetFileInfo(".env", ExcludedFiles.None)`, what should happen? IMO the override behavior is easier to reason about.

src/Microsoft.Extensions.FileProviders.Physical/Internal/PhysicalDirectoryContents.cs
```diff
@@ -17,19 +18,25 @@ public class PhysicalDirectoryContents : IDirectoryContents
...
 
-            _directory = directory;
+        /// <summary>
+        /// Initializes an instance of <see cref="PhysicalDirectoryContents"/>
+        /// </summary>
+        /// <param name="directory">The directory</param>
+        /// <param name="excludedFiles">Specifies which files are excluded from the directory listing.</param>
+        public PhysicalDirectoryContents(string directory, ExcludedFiles excludedFiles)
+        {
+            _directory = directory ?? throw new ArgumentNullException(nameof(directory));
```
natemcmaster: They have Fowler's (relucant) approval. I'll conceded throw expressions are not always the right choice, but I think they're a win for simple ctor checks.

src/Microsoft.Extensions.FileProviders.Physical/ExcludedFiles.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Exclude files and directories when the name begins with a period, or has either <see cref="FileAttributes.Hidden"/> or <see cref="FileAttributes.System"/> is set on <see cref="FileSystemInfo.Attributes"/>.
```
natemcmaster: Will update

src/Microsoft.Extensions.FileProviders.Physical/Internal/FileSystemInfoHelper.cs
natemcmaster: Will update.

natemcmaster:
Attempt 4 at picking good names
```c#
    [Flags]
    public enum ExclusionFilters
    {
        Sensitive = DotPrefixed | Hidden | System,
        DotPrefixed = 0x0001,
        Hidden = 0x0002,
        System = 0x0004,
        None = 0
    }
```

natemcmaster:
:up: :date: thanks for all the feedback so far. Here is the public API added in latest draft.

image
image

build/dependencies.props
pranavkm: That was easy: https://github.com/aspnet/FileSystem/pull/281

test/Microsoft.Extensions.FileProviders.Physical.Tests/ExcludedFilesTests.cs
pranavkm: The `excluded` parameter is essentially an if and `if`s in a test are generally an indication of a conflating different scenarios.

natemcmaster:
Per in person discussion with @glennc and @Tratcher, we won't overload methods, just constructors. This scopes back this feature but still unblocks those that want to get access to 'hidden' files through this abstraction.

src/Microsoft.Extensions.FileProviders.Physical/ExcludedFiles.cs
pranavkm: That sounds like a good name.

natemcmaster:
:bell: now that we've branched for 2.1, @Tratcher are you okay with this API addition?

src/Microsoft.Extensions.FileProviders.Physical/Internal/FileSystemInfoHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+using System.IO;
+
+namespace Microsoft.Extensions.FileProviders.Physical.Internal
+{
+    /// <summary>
+    /// Helpful methods for FileSystem APIs.
+    /// </summary>
+    public static class FileSystemInfoHelper
```
Eilon: Can this type stay `internal`? We kind of mostly regret having so many 'pub-ternal' types, so we'd rather not add too many more...

natemcmaster: Sure. It's not meant to be used by other components.

Eilon: Yeah that would be good... less future pain for us. Can you send an update for that? (New PR of course)

natemcmaster: will do

evil-shrike:
Hi. I as many others encountered shortcomings in PhysicalFileProvider in 2.0 and had to create my own implementation for allowing serving dot-files. 
Let me suggest more enhancements to `PhysicalFileProvider`.

* enum flags `ExclusionFilters` can be fine but wouldn't be safer to add an options object to ctor signature to allow future feature additions:
```
public PhysicalFileProvider(string root, ExclusionFilters filters)
```
what if it will be needed to add something more? Another argument?
I suggest creating `PhysicalFileProviderOptions`:
```
public class PhysicalFileProviderOptions 
{
    ExclusionFilters ExclusionFilters { get;set }
}
```
and use it as ctor argument instead:
```
public PhysicalFileProvider(string root, PhysicalFileProviderOptions options)
```
* Besides flags for exclusion modes I'd suggest adding Glob-patttens
```
public class PhysicalFileProviderOptions 
{
    ExclusionFilters ExclusionFilters { get;set }
    string[] ExcludePaths { get;set }
}
```
usage:
```
new PhysicalFileProvider("root", new PhysicalFileProviderOptions { 
  // I need include dot-files but exclude others (not very handy btw)
  ExclusionFilters = ExclusionFilters.Sensitive & ~ExclusionFilters.DotPrefixed ,
  // exclude ".idea" subfolder with all contentl; exclude *.map files indide 'src' folder
  ExcludePaths = new[] { ".idea/", "src/**/*.map" } 
});
```
I know you will say - use `StaticFileOptions.OnPrepareResponse` callback in StaticFilesMiddleware. But the same is appliable for dot-files - why not to filter them inside middleware instead of FileProvider?
* serving files with dots if it's just js/css is definitely normal, why are they excluded by default?
Moreover it's a breaking change as compared with IIS + ASP.NET 4.
So everybody have to write this expression over and over:
`ExclusionFilters.Sensitive & ~ExclusionFilters.DotPrefixed `
Suggestion - include dot-files by default (but exclude folders starting with ".")
* anyway of default, if we have the options object I'd suggest to move "exclude dot-files" option into a property:
```
public class PhysicalFileProviderOptions 
{
    ExclusionFilters ExcludeByAttributes { get;set }
    string[] Exclude { get;set }
    bool IncludeDotFile { get; set } // or ExcludeDotFiles
}
```
* make `GetFileInfo` and `GetDirectoryContents` methods virtual to allow overriding `PhysicalFileProvider`
* make PathUtils are public as currently it's hard to implement own PhysicalFileProvider with the similar logic - we have to copy-paste methods like PathNavigatesAboveRoot, HasInvalidPathChars and so on.
* Support special files inside folder like `.ignore` which would tell FileProvider to ignore the folder with all content inside. The file name would be passed via options object 
```
public class PhysicalFileProviderOptions 
{
    ExclusionFilters ExcludeByAttributes { get;set }
    string[] Exclude { get;set }
    bool IncludeDotFile { get; set } // or ExcludeDotFiles
    string IgnoreFileName {get;set;}
}
```

Tratcher:
@evil-shrike this PR was closed quite a while ago. Please open a new issue with your feedback.

PR #281, opened 2017-06-21
Update Moq version

pranavkm:


natemcmaster:
FYI @mikeharder - cref https://github.com/aspnet/Coherence-Signed/issues/583

mikeharder:
CC: @ryanbrandenburg 

Do we need to update this one repo before all the rest?  See https://github.com/aspnet/Coherence-Signed/issues/583.

For the Moq version, 4.7.49 is the earliest version with the fixes we need, but 4.7.58 and 4.7.63 have been released since.  Should we update to current latest 4.7.63?  I assume we should use a consistent version across all our projects unless there's a reason to do otherwise.

pranavkm:
They've been releasing a build rather frequently of late: https://www.nuget.org/packages/Moq. Might be difficult for us to keep up with the latest.

mikeharder:
Right now I would just use `4.7.49` since we know it fixes the NuGet warning and it has more downloads than the two newer releases.

Long-term I think we should have a plan for how to manage versions of all third-party dependencies.  A few options:

* Update only when needed
* Update on a regular schedule (1/month, 1/quarter, etc)
* Try to stay on latest version whenever it updates

pranavkm:
🆙 📅 to 4.7.49.

PR #289, opened 2017-07-31
Shorten folder names

natemcmaster:
image


MartinJohns:
Will this be a common approach across the Microsoft projects?

natemcmaster:
I cannot speak for all of Microsoft or even the .NET team, but within the small team that works on the aspnet core runtime, we intend to shorten folder names in many of our projects. We've been having consistent problems with paths extending beyond max path limits, so we're trying to mitigate it by shortening folder names.

FileSystem.sln
```diff
@@ -11,23 +11,23 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		NuGet.config = NuGet.config
 	EndProjectSection
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.FileProviders.Abstractions", "src\Microsoft.Extensions.FileProviders.Abstractions\Microsoft.Extensions.FileProviders.Abstractions.csproj", "{DD94B7E8-3A59-4F84-98A0-8139BE259A87}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.FileProviders.Abstractions", "src\Abstractions\Abstractions.csproj", "{DD94B7E8-3A59-4F84-98A0-8139BE259A87}"
```
Eilon: Per the other discussions these should probably all be `FS.Abstractions`, `FS.*` etc. Otherwise you end up with duplicate project filenames that are difficult to disambiguate.

natemcmaster: How would we get duplicate project filenames? If you mean between different projects, isn't the solution name and project folder enough to disambiguate?

natemcmaster:
:up: 📅 

image

FileSystem.sln
Eilon: I mean between different solutions. I agree there's no technical problem in that everything will always build fine, but rather about the filename that would show up in "Recently used files" in various places. Which "Project: Abstractions.csproj" do I click on if there's more than one?

natemcmaster: Ok, makes sense.

Issue #299, opened 2017-11-06
FileSystemGlobbing passes invalid path upon upgrade to Core 2.0


JeremyKuhne:
From https://github.com/dotnet/home/issues/88 opened by @HarelM:

``` C#
An unhandled exception of type 'System.ArgumentException' occurred in System.Private.CoreLib.dll
The path is not of a legal form.

System.Private.CoreLib.dll!System.IO.Path.GetFullPath(string path)	Unknown	No symbols loaded.
 	Microsoft.Extensions.FileSystemGlobbing.dll!Microsoft.Extensions.FileSystemGlobbing.InMemoryDirectoryInfo.InMemoryDirectoryInfo(string rootDir, System.Collections.Generic.IEnumerable<string> files, bool normalized)	Unknown	No symbols loaded.
 	Microsoft.Extensions.FileSystemGlobbing.dll!Microsoft.Extensions.FileSystemGlobbing.MatcherExtensions.Match(Microsoft.Extensions.FileSystemGlobbing.Matcher matcher, string rootDir, System.Collections.Generic.IEnumerable<string> files)	Unknown	No symbols loaded.
 	Microsoft.Extensions.FileSystemGlobbing.dll!Microsoft.Extensions.FileSystemGlobbing.MatcherExtensions.Match(Microsoft.Extensions.FileSystemGlobbing.Matcher matcher, string file)	Unknown	No symbols loaded.
 	Microsoft.Extensions.FileProviders.Physical.dll!Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.ReportChangeForMatchedEntries(string path)	Unknown	No symbols loaded.
 	Microsoft.Extensions.FileProviders.Physical.dll!Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.OnFileSystemEntryChange(string fullPath)	Unknown	No symbols loaded.
 	Microsoft.Extensions.FileProviders.Physical.dll!Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.OnRenamed(object sender, System.IO.RenamedEventArgs e)	Unknown	No symbols loaded.
 	System.IO.FileSystem.Watcher.dll!System.IO.FileSystemWatcher.NotifyRenameEventArgs(System.IO.WatcherChangeTypes action, string name, string oldName)	Unknown	No symbols loaded.
[...]

PR #305, opened 2017-12-12
Fix crash when OnFileSystemEntryChange is called with root path

yp87:
It is possible for `OnFileSystemEntryChange` to be called with root path. In that case, `ReportChangeForMatchedEntries` would be called with an empty string leading to a crash.

Addresses #299

### DIFF
--- a/src/FS.Physical/PhysicalFilesWatcher.cs
+++ b/src/FS.Physical/PhysicalFilesWatcher.cs
@@ -255,7 +255,11 @@ namespace Microsoft.Extensions.FileProviders.Physical
                 }
 
                 var relativePath = fullPath.Substring(_root.Length);
-                ReportChangeForMatchedEntries(relativePath);
+
+                if (relativePath != string.Empty)
+                {
+                    ReportChangeForMatchedEntries(relativePath);
+                }
             }
             catch (Exception ex) when (
                 ex is IOException ||


yp87:
I guess the failing unit test `TokenCallbackInvokedOnFileChange`
Should read:
`fileSystemWatcher.CallOnChanged(new FileSystemEventArgs(WatcherChangeTypes.Changed, fileLocation, fileName));`
instead of:
`fileSystemWatcher.CallOnChanged(new FileSystemEventArgs(WatcherChangeTypes.Changed, root.RootPath, fileName));`

yp87:
Seems like the unit test was failing before..

src/FS.Physical/PhysicalFilesWatcher.cs
```diff
@@ -255,7 +255,11 @@ private void OnFileSystemEntryChange(string fullPath)
                 }
 
                 var relativePath = fullPath.Substring(_root.Length);
-                ReportChangeForMatchedEntries(relativePath);
+
+                if (relativePath != string.Empty)
+                {
```
HarelM: Wouldn't it be better to use `!string.isNullOrWhiteSpace(...)` or `!string.isNullOrEmpty(...)`?

yp87: Could be but substring cannot return null per msdn.

yp87: Also, iirc, a file or folder cannot terminate with a whitespace, so a whitespace only relativePath should not happen.

HarelM: I can't argue with that logic, I personally just prefer to keep it on the safe side when it comes to string empty checks.

aspnet-hello:
This issue was moved to aspnet/Home#2534

aspnet-hello:
Sorry I was a bad bot. Re-opening this PR.

natemcmaster:
Thanks for taking a stab at this @yp87. This fix has been superceded by https://github.com/aspnet/FileSystem/pull/314, which is basically the same fix, but I added tests as well so we don't regress this. #299 should be fixed with this change. Please let me know if problems persist after updating to a version with this fix.

PR #308, opened 2017-12-14
Add a task to generate a manifest with the file system structure of the original files

javiercn:


src/FS.Embedded.Manifest.Task/GenerateEmbeddedResourcesManifest.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        private const string LogicalName = "LogicalName";
+        private const string ManifestResourceName = "ManifestResourceName";
+        private const string TargetPath = "TargetPath";
+
+        [Required]
+        public ITaskItem[] EmbeddedFiles { get; set; }
+
+        [Required]
+        public ITaskItem ManifestFile { get; set; }
```
natemcmaster: Seems like the only usage of this is to acquire the `TargetPath` metadata item, right? Any reason this couldn't just be `public string ManifestFilePath { get; set; } `?

test/FS.Embedded.Manifest.Task.Test/GenerateEmbeddedResourcesManifestTest.cs
```diff
@@ -0,0 +1,419 @@
...
+using System.Linq;
+using System.Text;
+using System.Xml;
+using System.Xml.Linq;
+using Microsoft.Build.Framework;
+using Xunit;
+
+namespace Microsoft.Extensions.FileProviders.Embedded.Manifest.Task
+{
+    public class GenerateEmbeddedResourcesManifestTest
```
natemcmaster: 👍 seems like good test coverage.

src/FS.Embedded.Manifest.Task/Entry.cs
```diff
@@ -0,0 +1,118 @@
...
+                throw new InvalidOperationException("Tried to add children to a file.");
+            }
+
+            var existing = Children.SingleOrDefault(c => c.HasName(child.Name));
+            if (existing != null && !existing.Equals(child))
+            {
+                throw new InvalidOperationException("Tried to add two different items with the same name");
+            }
+
+            Children.Add(child);
```
mkArtakMSFT: You don't have a check for a situation, where the child has already been added, I would throw in that case also, instead of silently ignoring the request

src/FS.Embedded.Manifest.Task/Entry.cs
```diff
@@ -0,0 +1,118 @@
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private class NameComparer : IComparer<Entry>
+        {
+            public static NameComparer Instance { get; set; } = new NameComparer();
```
mkArtakMSFT: Should the "set" be "private" to avoid reassignments?

src/FS.Embedded.Manifest.Task/GenerateEmbeddedResourcesManifest.cs
```diff
@@ -0,0 +1,91 @@
...
+            var processedItems = CreateEmbeddedItems(EmbeddedFiles);
+
+            var manifest = BuildManifest(processedItems);
+
+            var manifestItem = new EmbeddedItem
+            {
+                ManifestFilePath = ManifestFile.GetMetadata(TargetPath)
+            };
+
+            var document = manifest.ToXmlDocument();
```
mkArtakMSFT: Please move this inside the using scope

src/FS.Embedded.Manifest.Task/GenerateEmbeddedResourcesManifest.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        public EmbeddedItem[] CreateEmbeddedItems(params ITaskItem[] items) =>
+            items.Select(er => new EmbeddedItem
+            {
+                ManifestFilePath = GetManifestPath(er),
+                AssemblyResourceName = GetAssemblyResourceName(er)
+            }).ToArray();
+
+        string GetManifestPath(ITaskItem er) => string.Equals(er.GetMetadata(LogicalName), er.GetMetadata(ManifestResourceName)) ?
```
mkArtakMSFT: Please provide explicit visibility keywords for those methods

src/FS.Embedded.Manifest.Task/GenerateEmbeddedResourcesManifest.cs
```diff
@@ -0,0 +1,91 @@
...
+            NormalizePath(er.GetMetadata(LogicalName));
+
+        string GetAssemblyResourceName(ITaskItem er) => string.Equals(er.GetMetadata(LogicalName), er.GetMetadata(ManifestResourceName)) ?
+            er.GetMetadata(ManifestResourceName) :
+            er.GetMetadata(LogicalName);
+
+        string NormalizePath(string path) => Path.DirectorySeparatorChar == '\\' ?
+            path.Replace("/", "\\") : path.Replace("\\", "/");
+
+        public Manifest BuildManifest(EmbeddedItem[] processedItems)
```
mkArtakMSFT: validate the "processedItems" to be not null before usage

src/FS.Embedded.Manifest.Task/Manifest.cs
```diff
@@ -0,0 +1,75 @@
...
+            public static readonly string Name = "Name";
+            public static readonly string FileSystem = "FileSystem";
+            public static readonly string Root = "Manifest";
+            public static readonly string File = "File";
+            public static readonly string ResourcePath = "ResourcePath";
+        }
+
+        public Entry Root { get; set; } = Entry.Directory("");
+
+        public void AddElement(string originalPath, string assemblyResourceName)
```
mkArtakMSFT: Please validate arguments on all public methods

src/FS.Embedded.Manifest.Task/Entry.cs
```diff
@@ -0,0 +1,118 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.Extensions.FileProviders.Embedded.Manifest.Task
+{
+    [DebuggerDisplay("{Name,nq}")]
+    public class Entry : IEquatable<Entry>
```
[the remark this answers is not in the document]
javiercn: I can put it on the internal namespace, but I use this all over in testing and don't want to do internalsvisibleto dance.

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>Microsoft.Extensions.FileProviders.Embedded.Manifest.Task</RootNamespace>
```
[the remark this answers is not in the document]
javiercn: You mean the MSBuild part of it? (The targets file). I would definitely keep the assembly separate

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <AssemblyName>Microsoft.Extensions.FileProviders.Embedded.Manifest.Task</AssemblyName>
+    <Description>MSBuild task to generate a manifest that can be used by Microsoft.Extensions.FileProviders.Embedded to preserve
+    metadata of the files embedded in the assembly at compilation time.</Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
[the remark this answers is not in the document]
javiercn: Building on VS worked just fine and I don't think the dependency on 15.4 will be an issue.

src/FS.Embedded.Manifest.Task/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.Manifest.Task.targets
```diff
@@ -0,0 +1,39 @@
...
+
+  <ItemGroup Condition="$(EmbeddedFilesManifest) == 'true'">
+    <EmbeddedResource
+      Include="$(IntermediateOutputPath)$(EmbeddedFilesManifestFileName)"
+      IsManifest="true"
+      Condition="@(EmbeddedResource->WithMetadataValue('IsManifest','true')->Count()) == 0"
+      LogicalName="$(EmbeddedFilesManifestFileName)" />
+  </ItemGroup>
+
+  <Target Name="GenerateEmbeddedFilesManifest" AfterTargets="PrepareResourceNames" Condition="$(EmbeddedFilesManifest) == 'true'">
```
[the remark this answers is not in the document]
javiercn: As the supreme MSBuild incantator, can you tell me how to do this? The manifest is already order resilient

src/FS.Embedded.Manifest.Task/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.Manifest.Task.targets
```diff
@@ -0,0 +1,39 @@
...
+    <ItemGroup>
+      <__FilesForManifest Include="@(EmbeddedResource)" />
+      <__FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('ExcludeFromManifest','true'))" />
+      <__FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('IsManifest','true'))" />
+      <__ManifestFiles Include="@(EmbeddedResource->WithMetadataValue('IsManifest', 'true'))" />
+    </ItemGroup>
+
+    <Error Condition="@(__ManifestFiles->Count()) > 1" Text="Multiple manifest files found. Only one manifest file per assembly is allowed: %(__ManifestFiles.TargetPath)" />
```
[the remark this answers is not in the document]
javiercn: Target path is set by the default manifest above to be $(IntermediateOutputPath)$(EmbeddedFilesManifestName), someone can use their own EmbeddedResource item if they want to generate the manifest somewhere else for any reason.

javiercn:
Addressed all feedback /cc: @natemcmaster @mkArtakMSFT

src/FS.Embedded.Manifest.Task/GenerateEmbeddedResourcesManifest.cs
```diff
@@ -0,0 +1,91 @@
...
+            NormalizePath(er.GetMetadata(LogicalName));
+
+        private string GetAssemblyResourceName(ITaskItem er) => string.Equals(er.GetMetadata(LogicalName), er.GetMetadata(ManifestResourceName)) ?
+            er.GetMetadata(ManifestResourceName) :
+            er.GetMetadata(LogicalName);
+
+        private string NormalizePath(string path) => Path.DirectorySeparatorChar == '\\' ?
+            path.Replace("/", "\\") : path.Replace("\\", "/");
+
+        public Manifest BuildManifest(EmbeddedItem[] processedItems)
```
mkArtakMSFT: Minor: would be nice to have all the public methods before private ones (or private methods in the bottom)

src/FS.Embedded.Manifest.Task/Manifest.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Microsoft.Extensions.FileProviders.Embedded.Manifest.Task
+{
+    public class Manifest
+    {
+        public Entry Root { get; set; } = Entry.Directory("");
+
+        public void AddElement(string originalPath, string assemblyResourceName)
+        {
+            var paths = originalPath.Split(Path.DirectorySeparatorChar);
```
mkArtakMSFT: You're missing argument validation here. originalPath can be null and that'll result in NullReferenceException

src/FS.Embedded.Manifest.Task/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.Manifest.Task.targets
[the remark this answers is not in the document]
pranavkm: The convention is to have targets that are implementation details (private) to start with an underscore. Consider splitting this off in to two targets

```
<Target Name="_GenerateEmbeddedFilesManifest" AfterTargets="PrepareResourceNames" DependsOnTargets="GenerateEmbeddedFilesManifest" Condition="$(EmbeddedFilesManifest) == 'true'" />

<Target Name="GenerateEmbeddedFilesManifest">
...
</Target>
```
so a user could wire up the second one.

test/FS.Embedded.Manifest.Task.Test/FS.Embedded.Manifest.Task.Test.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
pranavkm: Desktop?

src/FS.Embedded.Manifest.Task/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.Manifest.Task.targets
```diff
@@ -0,0 +1,44 @@
...
+  <ItemGroup Condition="$(EmbeddedFilesManifest) == 'true'">
+    <EmbeddedResource
+      Include="$(IntermediateOutputPath)$(EmbeddedFilesManifestFileName)"
+      IsManifest="true"
+      Condition="@(EmbeddedResource->WithMetadataValue('IsManifest','true')->Count()) == 0"
+      LogicalName="$(EmbeddedFilesManifestFileName)" />
+  </ItemGroup>
+
+  <Target 
+    Name="GenerateEmbeddedFilesManifest"
```
pranavkm: Consider naming this `_GenerateEmbeddedFilesManifest` if you want this target to be an implementation detail and not "public"

src/FS.Embedded.Manifest.Task/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.Manifest.Task.targets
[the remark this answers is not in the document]
pranavkm: You want to use `@(__ManifestFiles)`. The current approach would stop after printing the first item.

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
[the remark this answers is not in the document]
natemcmaster: If we expect users to have 15.4 or newer, we should find a way to encode that expectation, such as setting the minimum nuget client version that corresponds to VS 15.4

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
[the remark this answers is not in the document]
natemcmaster: The assembly definitely needs to be separate. One way to layout this package is like this:
```
- Microsoft.Extensions.FileProviders.Embedded.nupkg
  - lib/
     - netstandard2.0/
        + Microsoft.Extensions.FileProviders.Embedded.dll
  - build/
     - netstandard2.0/
         + Microsoft.Extensions.FileProviders.Embedded.targets
         + Microsoft.Extensions.FileProviders.Embedded.Manifest.Task.dll
```

src/FS.Embedded.Manifest.Task/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.Manifest.Task.targets
```diff
@@ -0,0 +1,44 @@
...
+    AfterTargets="PrepareResourceNames"
+    Condition="$(EmbeddedFilesManifest) == 'true'"
+    Inputs="@(EmbeddedResource)"
```
pranavkm: Unfortunately this is going to run in to https://github.com/Microsoft/msbuild/issues/2470. csc works around this by manually tracking file hashes - https://github.com/Microsoft/msbuild/pull/1328/files

src/FS.Embedded.Manifest.Task/Entry.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            return null;
+        }
+
+        private bool HasName(string currentSegment)
+        {
+            return string.Equals(Name, currentSegment, StringComparison.Ordinal);
+        }
+
+        public bool Equals(Entry other)
```
pranavkm: `GetHashCode` \ `Equals(object)` overrides?

src/FS.Embedded.Manifest.Task/Entry.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+            else
+            {
+                return SameChildren(Children, other.Children);
+            }
+        }
+
+        private bool SameChildren(ISet<Entry> left, ISet<Entry> right)
+        {
+            if (left.Count != right.Count)
```
pranavkm: Isn't there a SetEquals on set?

src/FS.Embedded.Manifest.Task/GenerateEmbeddedResourcesManifest.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        public EmbeddedItem[] CreateEmbeddedItems(params ITaskItem[] items) =>
+            items.Select(er => new EmbeddedItem
+            {
+                ManifestFilePath = GetManifestPath(er),
+                AssemblyResourceName = GetAssemblyResourceName(er)
+            }).ToArray();
+
+        private string GetManifestPath(ITaskItem er) => string.Equals(er.GetMetadata(LogicalName), er.GetMetadata(ManifestResourceName)) ?
```
pranavkm: gross

src/FS.Embedded.Manifest.Task/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.Manifest.Task.targets
natemcmaster: This is a good start. You need a few additional things to make this work correctly. Inputs/Outputs only compares the timestamps of each file and asserts that `Inputs.Max(f => f.LastModifiedTime) < Outputs.Min(f =>f .LastModifiedTime)`. What it doesn't do is compare if the list of Inputs and Outputs changes between runs. So if I remove an embedded resource, the assertion still checks out, and the manifest isn't re-built, but the list of embedded files is wrong.

The way we've worked around this in dotnet/sdk is to add an additional input, which generates a hash of all items and relevant metadata.

For a good example of this, checkout the assembly attribute generation targets: https://github.com/dotnet/sdk/blob/2995c8495082d1dd4bcad35c37007375edf45c6f/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets#L114

```xml 
  <!-- 
    To allow version changes to be respected on incremental builds (e.g. through CLI parameters),
    create a hash of all assembly attributes so that the cache file will change with the calculated
    assembly attribute values and msbuild will then execute CoreGenerateAssembly to generate a new file.
  -->
  <Target Name="CreateGeneratedAssemblyInfoInputsCacheFile"
          DependsOnTargets="GetAssemblyAttributes">
    <PropertyGroup>
      <GeneratedAssemblyInfoInputsCacheFile>$(IntermediateOutputPath)$(MSBuildProjectName).AssemblyInfoInputs.cache</GeneratedAssemblyInfoInputsCacheFile>
    </PropertyGroup>

    <!-- We only use up to _Parameter1 for most attributes, but other targets may add additional assembly attributes with multiple parameters. -->
    <Hash ItemsToHash="@(AssemblyAttribute->'%(Identity)%(_Parameter1)%(_Parameter2)%(_Parameter3)%(_Parameter4)%(_Parameter5)%(_Parameter6)%(_Parameter7)%(_Parameter8)')">
      <Output TaskParameter="HashResult" PropertyName="_AssemblyAttributesHash" />
    </Hash>

    <WriteLinesToFile Lines="$(_AssemblyAttributesHash)" File="$(GeneratedAssemblyInfoInputsCacheFile)" Overwrite="True" WriteOnlyWhenDifferent="True" />
    
    <ItemGroup>
      <FileWrites Include="$(GeneratedAssemblyInfoInputsCacheFile)" />
    </ItemGroup>
  </Target>

  <Target Name="CoreGenerateAssemblyInfo"
          Condition="'$(Language)'=='VB' or '$(Language)'=='C#'"
          DependsOnTargets="CreateGeneratedAssemblyInfoInputsCacheFile"
          Inputs="$(GeneratedAssemblyInfoInputsCacheFile)"
          Outputs="$(GeneratedAssemblyInfoFile)">
    <ItemGroup>
      <!-- Ensure the generated assemblyinfo file is not already part of the Compile sources, as a workaround for https://github.com/dotnet/sdk/issues/114 -->
      <Compile Remove="$(GeneratedAssemblyInfoFile)" />
    </ItemGroup>

    <WriteCodeFragment AssemblyAttributes="@(AssemblyAttribute)" Language="$(Language)" OutputFile="$(GeneratedAssemblyInfoFile)">
      <Output TaskParameter="OutputFile" ItemName="Compile" />
      <Output TaskParameter="OutputFile" ItemName="FileWrites" />
    </WriteCodeFragment>
  </Target>
```

src/FS.Embedded.Manifest.Task/Entry.cs
javiercn: Equality for these sets is not based solely on the name. We check that we aren't adding two files with the same name but different metadata

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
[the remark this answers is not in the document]
javiercn: we'll discuss this offline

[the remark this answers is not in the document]
javiercn: The new packages layout will take care of this dependency

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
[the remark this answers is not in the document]
javiercn: I did this

src/FS.Embedded.Manifest.Task/Entry.cs
pranavkm: Yup. But you should be able to use ``SortedSet<Entry>.CreateSetComparer(EqualityComparer<Entry>.Default).Equals(left, right);`

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.targets
```diff
@@ -0,0 +1,61 @@
...
+      Include="$(IntermediateOutputPath)$(EmbeddedFilesManifestFileName)"
+      IsManifest="true"
+      Condition="@(EmbeddedResource->WithMetadataValue('IsManifest','true')->Count()) == 0"
+      LogicalName="$(EmbeddedFilesManifestFileName)" />
+  </ItemGroup>
+
+  <Target Name="CreateGeneratedManifestInfoInputsCacheFile">
+    <PropertyGroup>
+      <GeneratedManifestInfoInputsCacheFile>$(IntermediateOutputPath)$(MSBuildProjectName).EmbeddedFilesManifest.cache</GeneratedManifestInfoInputsCacheFile>    
+  </PropertyGroup>
```
pranavkm: Formatting

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.targets
```diff
@@ -0,0 +1,61 @@
...
+
+    <Message Importance="normal" Condition="'@(__FilesForManifest->Count())' == 0" Text="All embedded files are excluded or a manifest file. No embedded files found to include in the manifest"/>
+
+    <GenerateEmbeddedResourcesManifest
+      EmbeddedFiles="@(__FilesForManifest)"
+      ManifestFile="@(__ManifestFiles)"
+      Condition="@(__FilesForManifest->Count()) != 0" />
+
+  </Target>
+</Project>
```
pranavkm: new line at eof

NuGetPackageVerifier.json
```diff
@@ -2,6 +2,15 @@
     "Default": { // Rules to run for packages not listed in any other set.
         "rules": [
             "DefaultCompositeRule"
-        ]
+        ],
+        "packages": {
+            "Microsoft.Extensions.FileProviders.Embedded.Manifest.Task": {
```
natemcmaster: You can remove this now.

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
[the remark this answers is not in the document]
natemcmaster: 👍

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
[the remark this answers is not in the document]
natemcmaster: We still need to target .NET Framework. The package that uses this task would install in a project that could be as old as MSBuild 15.3. If you don't, VS users not on latest will hit TypeLoadException errors when they try to compile their app.

See https://github.com/Microsoft/msbuild/issues/2738

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.targets
```diff
@@ -0,0 +1,61 @@
+<Project>
+
+  <UsingTask
+    AssemblyFile="Microsoft.Extensions.FileSystem.Embedded.Manifest.Task"
```
natemcmaster: Needs to be a file path

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.targets
```diff
@@ -0,0 +1,61 @@
...
+    Inputs="$(GeneratedManifestInfoInputsCacheFile)"
+    Outputs="@(EmbeddedResource->WithMetadataValue('IsManifest', 'true'))">
+    <ItemGroup>
+      <__FilesForManifest Include="@(EmbeddedResource)" />
+      <__FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('ExcludeFromManifest','true'))" />
+      <__FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('IsManifest','true'))" />
+      <__ManifestFiles Include="@(EmbeddedResource->WithMetadataValue('IsManifest', 'true'))" />
+    </ItemGroup>
+
+    <Error Condition="'@(__ManifestFiles->Count())' > 1" Text="Multiple manifest files found. Only one manifest file per assembly is allowed: @(__ManifestFiles->TargetPath)" />
```
natemcmaster: `@(__ManifestFiles->TargetPath)` is this valid MSBuild syntax? I don't recognized it.

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.targets
```diff
@@ -0,0 +1,61 @@
...
+  <PropertyGroup>
+    <GenerateEmbeddedFilesManifest Condition="'$(GenerateEmbeddedFilesManifest)' == ''">false</GenerateEmbeddedFilesManifest>
+    <EmbeddedFilesManifestFileName Condition="'$(EmbeddedFilesManifestFileName)' == ''">Microsoft.Extensions.FileSystem.Embedded.Manifest.xml</EmbeddedFilesManifestFileName>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(GenerateEmbeddedFilesManifest)' == 'true'">
+    <EmbeddedResource
+      Include="$(IntermediateOutputPath)$(EmbeddedFilesManifestFileName)"
+      IsManifest="true"
+      Condition="@(EmbeddedResource->WithMetadataValue('IsManifest','true')->Count()) == 0"
```
natemcmaster: I'm not sure a users would ever need to write their own manifest. Seems better to always let our targets perform the manifest generation.

But if this is an important scenario, you should add a functional test that ensures this scenario works. From what I can tell, the GenerateEmbeddedResourcesManifest task would end up overwriting their file with generated code.

natemcmaster:
Example automated test for asserting incremental compliation: https://github.com/dotnet/sdk/blob/73fb22a07091585380ead0bcbcdc38839a0266e8/test/Microsoft.NET.Build.Tests/GivenThatWeWantToControlGeneratedAssemblyInfo.cs#L122-L194

src/FS.Embedded.Manifest.Task/FS.Embedded.Manifest.Task.csproj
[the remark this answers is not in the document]
natemcmaster: Example
```xml
  <PropertyGroup>
    <TaskAssembly Condition=" '$(MSBuildRuntimeType)' == 'Core'">.\bin\Debug\netstandard1.6\GreetingTasks.dll</TaskAssembly>
    <TaskAssembly Condition=" '$(MSBuildRuntimeType)' != 'Core'">.\bin\Debug\net46\GreetingTasks.dll</TaskAssembly>
  </PropertyGroup>

  <UsingTask TaskName="MSBuildTasks.SayHello" AssemblyFile="$(TaskAssembly)" />
```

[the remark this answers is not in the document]
natemcmaster: See also http://www.natemcmaster.com/blog/2017/07/05/msbuild-task-in-nuget/#the-solution

javiercn:
I've fix a bug for adding tests for incremental compilation and I will be setting up a meeting to discuss what approach we want to take here. https://github.com/aspnet/FileSystem/issues/309

natemcmaster:
:shipit: this feature's gonna be dope.

PR #310, opened 2017-12-18
Adds an embedded file provider that relies on an embedded manifest for resolving files.

javiercn:
* Implement manifest file provider
* Consumes the manifest from an embedded resource in the assembly.
* Allows the manifest file embedded resource name to be specified.
* Allows the file provider to expose only a subset of the file system in the manifest.


src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Embedded.Manifest
+{
+    internal class EmbeddedFilesManifest
+    {
+        private static readonly char[] _invalidFileNameChars = Path.GetInvalidFileNameChars()
+            .Where(c => c != Path.DirectorySeparatorChar && c != Path.AltDirectorySeparatorChar).ToArray();
+
+        private static readonly char[] Separators = new char[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
```
mkArtakMSFT: please be consistent in naming. fields should be lower case and start with an '_'.

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            _rootDirectory = rootDirectory;
+        }
+
+        internal ManifestEntry ResolveEntry(string path)
+        {
+            if (string.IsNullOrEmpty(path) || HasInvalidPathChars(path))
+            {
+                return null;
```
mkArtakMSFT: Should this throw instead?

mkArtakMSFT: You should throw instead.

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
```diff
@@ -0,0 +1,88 @@
...
+                return _rootDirectory;
+            }
+
+            var tokenizer = new StringTokenizer(trimmed, Separators);
+            ManifestEntry currentEntry = _rootDirectory;
+            foreach (var segment in tokenizer)
+            {
+                if (segment.Equals(""))
+                {
+                    return null;
```
mkArtakMSFT: Does this mean that the path was an empty string or something like "./"?

mkArtakMSFT: If so, shouldn't this return the _rootDirectory?

src/FS.Embedded/Manifest/ManifestDirectory.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            if (children == null)
+            {
+                throw new ArgumentNullException(nameof(children));
+            }
+
+            Children = children;
+        }
+
+        public ManifestEntry[] Children { get; protected set; }
```
mkArtakMSFT: Do you expect changes? My point is to use an IEnumerable<ManifestEntry> or IReadOnlyList<ManifestEntry> instead

src/FS.Embedded/Manifest/ManifestDirectoryInfo.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            Assembly = assembly;
+            Directory = directory;
+        }
+
+        public bool Exists => true;
+
+        public long Length => -1;
+
+        public string PhysicalPath => null;
```
mkArtakMSFT: How is this set?

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
```diff
@@ -0,0 +1,88 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.FileProviders.Embedded.Manifest
+{
+    internal class EmbeddedFilesManifest
+    {
+        private static readonly char[] _invalidFileNameChars = Path.GetInvalidFileNameChars()
```
natemcmaster: Looks similar to the result of  `Path.GetInvalidPathChars()`. Could you just use that instead?

src/FS.Embedded/Manifest/ManifestParser.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            var stream = assembly.GetManifestResourceStream(name);
+            if (stream == null)
+            {
+                throw new InvalidOperationException($"Can't find a manifest with name '{name}'");
```
natemcmaster: Suggested wording:
```
Could not load the embedded file manifest '{manfiestName}' for assembly '{assemblyName}'.
```

Seems like this will be the error message most users hit first. You might also consider adding info to the error about how to enable manifest generation.

src/FS.Embedded/Manifest/ManifestParser.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+
+            var document = XDocument.Load(stream);
+
+            var manifest = EnsureElement(document, "Manifest");
+            var manifestVersion = EnsureElement(manifest, "ManifestVersion");
+            var version = EnsureText(manifestVersion);
+            if (!string.Equals("1.0", version))
+            {
+                throw new InvalidOperationException($"Invalid manifest version: '{version}'");
```
natemcmaster: Add more context to the error message. 
Suggested wording:
```
The embedded file manifest '{manfiestName}' for assembly '{assemblyName}' specifies an unsupported file format version: '{version}'.
```

Also, how would a user go about resolving this?

src/FS.Embedded/Manifest/ManifestParser.cs
```diff
@@ -0,0 +1,160 @@
...
+            var entries = fileSystem.Elements();
+            var entriesList = new List<ManifestEntry>();
+            foreach (var element in entries)
+            {
+                try
+                {
+                    var entry = BuildEntry(element);
+                    entriesList.Add(entry);
+                }
+                catch (InsufficientExecutionStackException e)
```
natemcmaster: IIRC this is an uncatchable exception, similar to StackOverflowException

test/FS.Embedded.Tests/Manifest/ManifestEntryTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+using System;
+using Xunit;
+
+namespace Microsoft.Extensions.FileProviders.Embedded.Manifest
+{
+    public class ManifestEntryTests
+    {
+        [Fact]
+        public void TraversingAFile_ReturnsUnkownPath()
```
natemcmaster: Typo: Unknown

test/FS.Embedded.Tests/Manifest/ManifestEntryTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        public void TraversingWithDot_ReturnsSelf()
+        {
+            // Arrange
+            var directory = ManifestDirectory.CreateDirectory("a", Array.Empty<ManifestEntry>());
+
+            // Act
+            var result = directory.Traverse(".");
+
+            // Assert
+            Assert.Equal(directory, result);
```
natemcmaster: Assert.Same? These should be the same instance, right?

test/FS.Embedded.Tests/Manifest/ManifestParserTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            // Act
+            var manifest = ManifestParser.Parse(assembly, "Manifest.xml");
+
+            // Assert
+            Assert.NotNull(manifest);
+        }
+
+        [Fact]
+        public void Parse_ThrowsForEntriesWithDifferentCasing()
```
natemcmaster: Does the manifest generation task issue warnings or errors for this scenario? If it doesn't, consider adding this so we can catch this early

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
javiercn: Null signals a malformed path. The general guideline with these types is to use a special value instead of throwing:
See:
https://github.com/aspnet/FileSystem/pull/310/files#diff-498192ac768bf12b9b2d56daea3ff201R72
https://github.com/aspnet/FileSystem/pull/310/files#diff-498192ac768bf12b9b2d56daea3ff201R91

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
javiercn: This means that there were two separators together on the path, like "//" or like "wwwroot//site.css"
See:
https://github.com/aspnet/FileSystem/pull/310/files#diff-841c910febd853df519b2ff09175ed0cR35
https://github.com/aspnet/FileSystem/pull/310/files#diff-841c910febd853df519b2ff09175ed0cR13

javiercn: ./ will return _rootDirectory

src/FS.Embedded/Manifest/ManifestDirectoryInfo.cs
javiercn: It is never set. See https://github.com/aspnet/FileSystem/blob/dev/src/FS.Abstractions/IFileInfo.cs#L25

src/FS.Embedded/Manifest/ManifestParser.cs
javiercn: This issue will be cause most likely if we ever update the format of the manifest in a future release in a way that forces us to rev the version.
This will would be caused by a library compiled against a newer version of this package, being consumed by an older version of this assembly.
I'm not sure how to word that succinctly and if it would be valuable to do so.

Do you have any suggestion?

test/FS.Embedded.Tests/Manifest/ManifestParserTests.cs
javiercn: I've thought of this but decided to leave it for now for a few reasons:
* I don't want the manifest to be opinionated.
* We do case-insensitivity by default because that prevents issues with running on windows and failing on Linux.
* Its not very common to have files in your project (content) that only differ on casing.
* We might need to support case sensitivity in the future for things like static files once we have a clearer picture of how this will be used in the context of razor in class libraries. (We explicitly excluded providing this option for preview, but it might make a come back.)

src/FS.Embedded/Manifest/ManifestRootDirectory.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    internal class ManifestRootDirectory : ManifestDirectory
+    {
+        public ManifestRootDirectory(ManifestEntry[] children)
+            : base(null, children)
+        {
+            SetParent(ManifestSinkDirectory.Instance);
+        }
+
+        public override ManifestDirectory ToRootDirectory() => this;
```
pranavkm: Does this type need to account for path navigation outside the root?

src/FS.Embedded/Manifest/ManifestDirectoryInfo.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    return _lastModified.Value;
+                }
+                _lastModified = DateTimeOffset.UtcNow;
+
+                if (!string.IsNullOrEmpty(Assembly.Location))
+                {
+                    try
+                    {
+                        _lastModified = File.GetLastWriteTimeUtc(Assembly.Location);
```
pranavkm: Can we read these values once and pass it through (they're not meant to change once the App loads)?

src/FS.Embedded/Manifest/ManifestDirectoryContents.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    yield break;
+                }
+
+                foreach (var entry in Directory.Children)
+                {
+                    switch (entry)
+                    {
+                        case ManifestFile f:
+                            yield return new ManifestFileInfo(Assembly, f);
```
pranavkm: Can we cache these? vanilla `EmbeddedFileProvider` did that I think and there's no reason we can't do the same here.

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
javiercn: This follows whatever physicalfileprovider does, I could use what you are suggesting, but I don't know if there is a good reason for it. Maybe @Tratcher knows? See https://github.com/aspnet/FileSystem/blob/dev/src/FS.Physical/Internal/PathUtils.cs#L12

src/FS.Embedded/Manifest/ManifestDirectoryInfo.cs
javiercn: I'm still not happy with this behavior of reading the assembly. I'm giving it some thought

src/FS.Embedded/Manifest/ManifestRootDirectory.cs
javiercn: It does automatically by virtue of the parent being a sink directory

src/FS.Embedded/Manifest/ManifestParser.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+
+            return value;
+        }
+
+        private static string EnsureText(XElement element)
+        {
+            if (element.Elements().Count() == 0 &&
+                !element.IsEmpty &&
+                element.Nodes().Count() == 1 &&
```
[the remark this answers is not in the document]
javiercn: Some of my tests didn't like this suggestion.

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
Tratcher: I don't. That code has been refactored a few times since I looked at it last.

test/FS.Embedded.Tests/Manifest/ManifestEntryTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        [Fact]
+        public void TraversingRootDirectoryWithDotDot_ReturnsSinkDirectory()
+        {
+            // Arrange
+            var directory = ManifestDirectory.CreateRootDirectory(Array.Empty<ManifestEntry>());
+
+            // Act
+            var result = directory.Traverse("..");
```
pranavkm: One more test for `..\..`, plus ones where we `CreateRootDirectory()` and then attempt to navigate past root.

src/FS.Embedded/Manifest/ManifestDirectoryContents.cs
javiercn: I think a low level component like this shouldn't care about caching and let that for something higher level. I won't do anything right now as we don't have any data that suggest this would have a terrible impact on perf. If we discover that is causing trouble in the future, we can just cache it at the ManifestFileProvider level. (Note that there is some caching going on already on the enumeration)

test/FS.Embedded.Tests/Manifest/ManifestEntryTests.cs
javiercn: Those tests are covered by TraversingUnknownPath_ReturnsSinkDirectory unless i'm not understanding something

src/FS.Embedded/Manifest/EmbeddedFilesManifest.cs
javiercn:
```
Path.GetInvalidFileNameChars().Except(Path.GetInvalidPathChars())
ExceptIterator { ':', '*', '?', '\\', '/' }
```
So I think this should stay as it is.

javiercn:
Addressed all the feedback

pranavkm:
Can you redo your base branch? It currently includes changes from your other PR.

test/FS.Embedded.Tests/Manifest/ManifestEntryTests.cs
pranavkm: More along the lines of 

```
var directory = ManifestDirectory.CreateRootDirectory(
  ManifestDirectory.CreateDirectory("a", "test1.txt"),
  ManifestDirectory.CreateDirectory("b", "test2.txt"));

var newRoot = directory.Traverse("a").ToRootDirectory();
Assert.Same(UnknownPath, newRoot.Traverse("../b/test.txt"));
```

src/FS.Embedded/Manifest/ManifestParser.cs
```diff
@@ -0,0 +1,159 @@
...
+                throw new InvalidOperationException($"Could not load the embedded file manifest " +
+                    $"'{name}' for assembly '{assembly.GetName().Name}'.");
+            }
+
+            var document = XDocument.Load(stream);
+
+            var manifest = EnsureElement(document, "Manifest");
+            var manifestVersion = EnsureElement(manifest, "ManifestVersion");
+            var version = EnsureText(manifestVersion);
+            if (!string.Equals("1.0", version))
```
pranavkm: `StringComparison.Ordinal`

test/FS.Embedded.Tests/TestFileInfo.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+
+namespace Microsoft.Extensions.FileProviders
+{
+    internal class TestFileInfo : IFileInfo
+        {
```
pranavkm: Formatting

build/dependencies.props
```diff
@@ -4,13 +4,13 @@
   </PropertyGroup>
   <PropertyGroup Label="Package Versions">
     <InternalAspNetCoreSdkPackageVersion>2.1.0-preview1-15626</InternalAspNetCoreSdkPackageVersion>
-    <MicrosoftAspNetCoreTestingPackageVersion>2.1.0-preview1-27849</MicrosoftAspNetCoreTestingPackageVersion>
+    <MicrosoftAspNetCoreTestingPackageVersion>2.1.0-preview1-27854</MicrosoftAspNetCoreTestingPackageVersion>
```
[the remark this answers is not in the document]
javiercn: yep

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileSystem.Embedded.targets
```diff
@@ -0,0 +1,66 @@
+<Project>
```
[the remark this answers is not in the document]
javiercn: Probably whitespace clean-up

javiercn:
This was merged af6fa5abe701e62d6f3ef5e6ff336c5da293e5a8

PR #315, opened 2018-01-19
[2.0.x] handle flaky OnRenamed events more gracefully

natemcmaster:
Backports #314 to the 2.0.x release branch, with the addition of a quirks mode switch in the event this causes unforseen behavior issues.

Resolves aspnet/Home#2808

cc @Eilon

### DIFF
--- a/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
+++ b/src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
@@ -245,6 +245,14 @@ namespace Microsoft.Extensions.FileProviders.Physical
 
         private void ReportChangeForMatchedEntries(string path)
         {
+            if (string.IsNullOrEmpty(path))
+            {
+                // System.IO.FileSystemWatcher may trigger events that are missing the file name,
+                // which makes it appear as if the root directory is renamed or deleted. Moving the root directory
+                // of the file watcher is not supported, so this type of event is ignored.
+                return;
+            }
+
             path = NormalizePath(path);
 
             var matched = false;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
```diff
@@ -48,6 +56,7 @@ public class PhysicalFilesWatcher : IDisposable
             FileSystemWatcher fileSystemWatcher,
             bool pollForChanges)
         {
+            AppContext.TryGetSwitch("Switch.Microsoft.AspNetCore.FileProviders.Issue2808", out _issue2808QuirksMode);
```
Eilon: Please use a meaningful name for this, not a bug number. See https://github.com/aspnet/Home/wiki/Quirks-mode-switches-in-patches

natemcmaster: 🤔 https://github.com/aspnet/Microsoft.Data.Sqlite/pull/477/files

This one is hard to name. Any suggestions? Switch.Microsoft.AspNetCore.FileProviders.DoNotThrowArgumentExceptionForFlakyEvents is the most accurate name I can think of...though it's certainly not concise.

natemcmaster:
Thought about this more. Honestly, I'm not sure the quirks mode switch provides much value. I can't think of any scenario in which users would want to preserve the ArgumentException, which is always thrown when the argument given to `ReportChangeForMatchedEntries(string path)` is null or empty

src/Microsoft.Extensions.FileProviders.Physical/PhysicalFilesWatcher.cs
Eilon: `IgnoreEmptyRenameEvents`? The name doesn't have to be pretty, I just don't want them to look like some kind of hacker code... Imagine you're working on an app and see that switch name - what would you even think of it? What is `Issue2808`?

natemcmaster: Let's got with IgnoreEmptyRenameEvents. Issue2808 is aspnet/Home#2808, but there is doubt this would be helpful to users trying to reason about the switch name

natemcmaster:
Talked in person. Although we think there is a very low risk of this breaking anyone, we'll keep the quirks mode switch in the event there is some scenario we didn't think of.

PR #316, opened 2018-01-19
Use unique names for properties

pranavkm:
* Only add the generated manifest file if we decide to run the task
* Cleanup some names to follow MSBuild-isms
* Move user-settable properties to props file

Fixes https://github.com/aspnet/Home/issues/2811


src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileProviders.Embedded.targets
```diff
@@ -1,66 +1,80 @@
...
+    AssemblyFile="$(_FileProviderTaskAssembly)"
     TaskName="Microsoft.Extensions.FileProviders.Embedded.Manifest.Task.GenerateEmbeddedResourcesManifest" />
 
   <PropertyGroup>
-    <GenerateEmbeddedFilesManifest Condition="'$(GenerateEmbeddedFilesManifest)' == ''">false</GenerateEmbeddedFilesManifest>
-    <EmbeddedFilesManifestFileName Condition="'$(EmbeddedFilesManifestFileName)' == ''">Microsoft.Extensions.FileProviders.Embedded.Manifest.xml</EmbeddedFilesManifestFileName>
+    <PrepareResourceNamesDependsOn>_CalculateEmbeddedFilesManifestInputs;$(PrepareResourceNamesDependsOn)</PrepareResourceNamesDependsOn>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(GenerateEmbeddedFilesManifest)' == 'true'">
```
pranavkm: Doing this is problematic since we may not generate the file. I moved it to a separate target that runs before resx generation that only adds it if there are any manifest files of interest.

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileProviders.Embedded.targets
```diff
@@ -1,66 +1,80 @@
...
     </PropertyGroup>
 
-    <Hash ItemsToHash="@(EmbeddedResource)">
+    <ItemGroup>
+      <_FilesForManifest Include="@(EmbeddedResource)" />
+      <_FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('ExcludeFromManifest','true'))" />
+    </ItemGroup>
+
+    <Warning 
+      Text="GenerateEmbeddedFilesManifest was set, but no EmbeddedResource items were found that could be added to the manifest."
```
pranavkm: Changes the message to a warning.

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileProviders.Embedded.targets
```diff
@@ -1,66 +1,80 @@
...
-      <__FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('ExcludeFromManifest','true'))" />
-      <__FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('IsManifest','true'))" />
-      <__ManifestFiles Include="@(EmbeddedResource->WithMetadataValue('IsManifest', 'true'))" />
+      <!-- Rebuild _FilesForManifest since PrepareResourceNames would have updated EmbeddedResource. -->
+      <_FilesForManifest Remove="@(_FilesForManifest)" />
+      <_FilesForManifest Include="@(EmbeddedResource)" />
+      <_FilesForManifest Remove="@(EmbeddedResource->WithMetadataValue('ExcludeFromManifest','true'))" />
     </ItemGroup>
 
-    <Error Condition="'@(__ManifestFiles->Count())' > 1" Text="Multiple manifest files found. Only one manifest file per assembly is allowed: @(__ManifestFiles)" />
```
pranavkm: We don't need this since we use `_GeneratedManifestFile` to track the output.

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileProviders.Embedded.targets
```diff
@@ -1,66 +1,80 @@
 <Project>
 
   <PropertyGroup>
-    <TaskFolder Condition=" '$(MSBuildRuntimeType)' == 'Core' ">netstandard1.5</TaskFolder>
-    <TaskFolder Condition=" '$(MSBuildRuntimeType)' != 'Core' ">net461</TaskFolder>
-    <TaskAssembly>$(MSBuildThisFileDirectory)..\..\tasks\$(TaskFolder)\Microsoft.Extensions.FileProviders.Embedded.Manifest.Task.dll</TaskAssembly>
+    <_FileProviderTaskFolder Condition="'$(MSBuildRuntimeType)' == 'Core'">netstandard1.5</_FileProviderTaskFolder>
```
pranavkm: This is too generic a name. Razor has the same issue.

src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileProviders.Embedded.props
```diff
@@ -0,0 +1,6 @@
+<Project>
```
pranavkm: Moved this a props file so the csproj can access the name for instance if it chooses to.

src/FS.Embedded/FS.Embedded.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <None Include="build\netstandard2.0\*" Pack="true" PackagePath="build\netstandard2.0\" />
+    <None Include="buildMultiTargeting\*" Pack="true" PackagePath="buildMultiTargeting\" />
```
natemcmaster: Why are we multi targeting? Embedded resources only executes in the inner-build.

pranavkm: That's true. But I wanted to ensure configurable properties are always available to the csproj and not just in the inner build.

natemcmaster: I can't really think of a scenario for a user needing the default props from an outer-build, but I'll trust your judgement.

pranavkm: It just makes access to these properties consistent.

PR #325, opened 2018-06-13
Set 2.1 baselines

ryanbrandenburg:


src/FS.Physical/baseline.netcore.json
```diff
@@ -88,51 +88,67 @@
...
-          "Virtual": true,
-          "ImplementedInterface": "System.Collections.Generic.IEnumerable<Microsoft.Extensions.FileProviders.IFileInfo>",
-          "Visibility": "Public",
-          "GenericParameter": []
+          "GenericParameter": [],
+          "Literal": "1"
         },
         {
-          "Kind": "Constructor",
-          "Name": ".ctor",
```
ryanbrandenburg: The constructor which takes only a directory disappeared, is was this intentional?

ryanbrandenburg: Nevermind, that's just the diff getting scrambled.

src/FS.Globbing/baseline.netcore.json
```diff
@@ -473,1179 +473,6 @@
       ],
       "GenericParameters": []
     },
-    {
```
natemcmaster: Why are there 1300 line deletions here?

ryanbrandenburg: Universe sets ExcludeInternalsPublic to true in order to only compare public items, but that parameter also controls how the files are generated. since these baselines a copied from the 2.1 Universe build the pub-ternals got removed. I filed aspnet/BuildTools#693 about the issue. I can attempt to modify my script to generate these from the 2.1.1 tag instead of from Universe, or we can deal with it for this time and make sure aspnet/BuildTools#693 is done for next release. Which do you prefer?

natemcmaster: I'm neutral on this one. Just wanted to be sure we knew why it was happening.

PR #333, opened 2018-08-21
Allow file watcher to actively poll for changes

pranavkm:
The ChangeToken.OnChange pattern that's commonly used by providers to listen to
changes requires IChangeToken to be active. The only two instances in the framework
that does not support are the PollingChangeToken.

This change makes the polling be active unless configured otherwise.

Fixes https://github.com/aspnet/Mvc/issues/8173


sebastienros:
Should we also fix https://github.com/aspnet/Home/issues/3110 here ?

src/FS.Physical/PhysicalFileProvider.cs
```diff
@@ -64,26 +64,115 @@ private PhysicalFileProvider(string root, PhysicalFilesWatcher physicalFilesWatc
...
+        /// uses polling to determine file changes.
+        /// <para>
+        /// By default, <see cref="PhysicalFileProvider"/>  uses <see cref="FileSystemWatcher"/> to listen to file change events
+        /// for <see cref="Watch(string)"/>. <see cref="FileSystemWatcher"/> is ineffective in some scenarios such as mounted drives.
+        /// Polling is required to effectively watch for file changes.
+        /// </para>
+        /// <seealso cref="UseActivePolling"/>.
+        /// </summary>
+        /// <value>
+        /// Defaults to <see langword="true"/>, if the environment variable <c>DOTNET_USE_POLLING_FILE_WATCHER</c> is <c>true</c> or <c>1</c>.
```
sebastienros: Please change the wording so it's clear that the default is actually `false`, and it will take whatever value in the ENV otherwise.

src/FS.Physical/PhysicalFileProvider.cs
```diff
@@ -64,26 +64,115 @@ private PhysicalFileProvider(string root, PhysicalFilesWatcher physicalFilesWatc
...
+            _useActivePolling = pollForChanges;
         }
 
         /// <summary>
         /// Disposes the provider. Change tokens may not trigger after the provider is disposed.
         /// </summary>
         public void Dispose()
         {
-            _filesWatcher.Dispose();
+            _fileWatcher?.Dispose();
```
sebastienros: We also need a finalize (to fix the other issue)

src/FS.Physical/PhysicalFilesWatcher.cs
```diff
@@ -339,26 +374,56 @@ private static bool IsDirectoryPath(string path)
...
         {
-            if (matchInfo.TokenSource.IsCancellationRequested)
+            if (cancellationTokenSource.IsCancellationRequested)
             {
-                return;
+                return Task.CompletedTask;
             }
 
-            Task.Run(() =>
+            return Task.Run(() =>
```
davidfowl: Avoid the closure 😄

pranavkm: I'll undo the change. That said, not super sure how I'd avoid the closure here. Need to pass in a cts to a background Task

pranavkm:
🆙 📅

src/FS.Physical/PhysicalFilesWatcher.cs
davidfowl: Like this https://github.com/aspnet/Caching/blob/ee13d634568fc5348e09fc090fab40233b1107cc/src/Microsoft.Extensions.Caching.Memory/MemoryCache.cs#L297

pranavkm:
🆙 📅

PR #334, opened 2018-08-26
Don't capture AsyncLocals to PhysicalFilesWatcher

benaadams:
This causes a problem with `IHttpContextAccessor` as it captures the `HttpContext` and everything linked to it causing a lot of memory to become rooted (as seen in https://github.com/aspnet/KestrelHttpServer/issues/2840#issuecomment-416034872)

image

/cc @davidfowl

### DIFF
--- a/src/FS.Physical/PhysicalFilesWatcher.cs
+++ b/src/FS.Physical/PhysicalFilesWatcher.cs
@@ -373,8 +373,27 @@ namespace Microsoft.Extensions.FileProviders.Physical
                 if ((!_filePathTokenLookup.IsEmpty || !_wildcardTokenLookup.IsEmpty) &&
                     !_fileWatcher.EnableRaisingEvents)
                 {
-                    // Perf: Turn off the file monitoring if no files to monitor.
-                    _fileWatcher.EnableRaisingEvents = true;
+                    // Don't capture the current ExecutionContext and its AsyncLocals onto the events causing them to live forever
+                    var restoreFlow = false;
+                    try
+                    {
+                        if (!ExecutionContext.IsFlowSuppressed())
+                        {
+                            ExecutionContext.SuppressFlow();
+                            restoreFlow = true;
+                        }
+
+                        // Perf: Turn off the file monitoring if no files to monitor.
+                        _fileWatcher.EnableRaisingEvents = true;
+                    }
+                        finally
+                    {
+                        // Restore the current ExecutionContext
+                        if (restoreFlow)
+                        {
+                            ExecutionContext.RestoreFlow();
+                        }
+                    }
                 }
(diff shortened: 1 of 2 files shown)


src/FS.Physical/PhysicalFilesWatcher.cs
```diff
@@ -73,11 +73,28 @@ public class PhysicalFilesWatcher : IDisposable
             _root = root;
             _fileWatcher = fileSystemWatcher;
             _fileWatcher.IncludeSubdirectories = true;
-            _fileWatcher.Created += OnChanged;
-            _fileWatcher.Changed += OnChanged;
-            _fileWatcher.Renamed += OnRenamed;
-            _fileWatcher.Deleted += OnChanged;
-            _fileWatcher.Error += OnError;
+            // Don't capture the current ExecutionContext and its AsyncLocals onto the token registration causing them to live forever
```
benaadams: event registration

benaadams:
Will only capture on linux (due to cancellation token) issue on Windows is due to the ChangeTokens (addressed elsewhere)

benaadams:
For linux I raised an issue https://github.com/dotnet/corefx/issues/31973

However all the FileSystemWatcher roots on your trace should be fixed by https://github.com/aspnet/Common/pull/392

benaadams:
> @benaadams this should be fixed by the use of 32dbcac.

Different capturing; its captured onto the FileSystemWatcher when the events are enabled with `_fileWatcher.EnableRaisingEvents = true;` https://github.com/dotnet/corefx/issues/31973#issuecomment-424083196

Unfortunately I can't write a test for it as ASP.NET Core now registers the callback as a cancellation token with the flow suppressed https://github.com/aspnet/Common/pull/392; so while it captures it; it doesn't flow it back to be observable :-/

/cc @stephentoub

src/FS.Physical/PhysicalFilesWatcher.cs
```diff
@@ -373,8 +373,27 @@ private void TryEnableFileSystemWatcher()
...
+                    var restoreFlow = false;
+                    try
+                    {
+                        if (!ExecutionContext.IsFlowSuppressed())
+                        {
+                            ExecutionContext.SuppressFlow();
+                            restoreFlow = true;
+                        }
+
+                        // Perf: Turn off the file monitoring if no files to monitor.
```
benaadams: This comment is terrible

mkArtakMSFT:
Thank you for your effort. As part of the effort to reduce the number of repositories we're going to move the code of this repo to https://github.com/aspnet/aspnetcore repo. Feel free to resubmit a new PR in the target repo as we're closing this PR now.

PR #335, opened 2018-08-31
Simplify file watching

pranavkm:
* Disable FileSystemWatcher when polling is enabled.
* Let the PollingFileWatcher determine when polling change tokens need to refresh

### DIFF
--- a/src/FS.Physical/PhysicalFileProvider.cs
+++ b/src/FS.Physical/PhysicalFileProvider.cs
@@ -27,8 +27,8 @@ namespace Microsoft.Extensions.FileProviders
 
         private readonly ExclusionFilters _filters;
 
-        private readonly Func<PhysicalFilesWatcher> _fileWatcherFactory;
-        private PhysicalFilesWatcher _fileWatcher;
+        private readonly Func<IFileWatcher> _fileWatcherFactory;
+        private IFileWatcher _fileWatcher;
         private bool _fileWatcherInitialized;
         private object _fileWatcherLock = new object();
 
@@ -76,7 +76,6 @@ namespace Microsoft.Extensions.FileProviders
         /// for <see cref="Watch(string)"/>. <see cref="FileSystemWatcher"/> is ineffective in some scenarios such as mounted drives.
         /// Polling is required to effectively watch for file changes.
         /// </para>
-        /// <seealso cref="UseActivePolling"/>.
         /// </summary>
         /// <value>
         /// The default value of this property is determined by the value of environment variable named <c>DOTNET_USE_POLLING_FILE_WATCHER</c>.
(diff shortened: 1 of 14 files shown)


src/FS.Physical/PhysicalFilesWatcher.cs
```diff
@@ -21,12 +21,10 @@ namespace Microsoft.Extensions.FileProviders.Physical
     ///     Triggers events on <see cref="IChangeToken" /> when files are created, change, renamed, or deleted.
     ///     </para>
     /// </summary>
-    public class PhysicalFilesWatcher : IDisposable
+    internal sealed class PhysicalFilesWatcher : IFileWatcher
```
pranavkm: Breaking changes for 3.0

src/FS.Physical/PhysicalFileProvider.cs
```diff
@@ -101,37 +100,7 @@ public bool UsePollingFileWatcher
...
-        /// </para>
-        /// <para>
-        /// This property is only effective when <see cref="UsePollingFileWatcher"/> is set.
-        /// </para>
-        /// </summary>
-        /// <value>
-        /// The default value of this property is determined by the value of environment variable named <c>DOTNET_USE_POLLING_FILE_WATCHER</c>.
-        /// When <c>true</c> or <c>1</c>, this property defaults to <c>true</c>; otherwise false.
-        /// </value>
-        public bool UseActivePolling
```
pranavkm: I took these options out - we'll hopefully have ample time to sort bugs with the polling and Timer our before RTM. Supporting multiple modes just makes the code harder to work with.

src/FS.Physical/PhysicalFilesWatcher.cs
natemcmaster: Why?

natemcmaster (changes requested):
I'd like to see if we can make improvements without API breaking changes of this magnitude.

src/FS.Physical/PhysicalFilesWatcher.cs
davidfowl: Yea, we shouldn't do this.

pranavkm: Ok. I assumed these were all internal implementation details that we should haven't made public and 3.0 would be a fair opportunity to change it. That said, we would have to change it's behavior \ API. Currently this type supports polling (it's a ctor parameter).

pranavkm (commented):
🆙 📅

src/FS.Physical/PollingFileWatcher.cs
```diff
@@ -0,0 +1,120 @@
...
+        private readonly Timer _timer;
+
+        public PollingFileWatcher(string root, TimeSpan pollingInterval)
+        {
+            _root = root;
+            PollingChangeTokens = new ConcurrentDictionary<string, IPollingChangeToken>(StringComparer.Ordinal);
+
+            // PhysicalFileProvider lazy initializes IFileWatcher instances. It's OK for us to initialize this in the ctor since we anticipate
+            // GetFileChangeToken to be invoked right after.
+            _timer = new Timer(RaiseChangeEvents, state: PollingChangeTokens, dueTime: pollingInterval, period: pollingInterval);
```
benaadams: Don't capture asynclocals onto timer https://github.com/dotnet/corefx/issues/26523#issuecomment-417926130

pranavkm: Ok. I port the workaround to 2.2 too

pranavkm:
🆙 📅

mkArtakMSFT:
Thank you for your effort. As part of the effort to reduce the number of repositories we're going to move the code of this repo to https://github.com/aspnet/aspnetcore repo. Feel free to resubmit a new PR in the target repo as we're closing this PR now.

natemcmaster:
*https://github.com/aspnet/Extensions, not aspnet/AspNetCore.

PR #340, opened 2018-10-17
Update tasks for netstandard2.0

rynowak:
The 15.8 version of MSBuild dropped support for netstandard1.5

This is resolves a build break in UniverseCoherence


natemcmaster:
:shipit:

